PR #14, opened 2023-11-22
support aid mode

Leozaki:


src/main.c
```diff
@@ -262,35 +393,71 @@ static int entry_profile_disable(int argc, char **argv)
...
+        printf("Usage: %s profile delete <iccid/aid>\n", argv[0]);
         return -1;
     }
 
-    iccid = argv[3];
+    param = argv[3];
+    len = strlen(param);
 
-    if ((ret = es10c_delete_profile_iccid(&ctx, iccid)))
+    if (len == 19 || len == 20)
```
septs, line 408: ICCID digit length is 18 to 22 digit code

see https://www.emnify.com/iot-glossary/iccid-number

PR #23, opened 2023-12-15
feat: $PWD independent executable file

CoelacanthusHex:
- Use RPATH on Linux/macOS to ensure it will search directory that contains lpac executable file.
- Windows doesnn't has RPATH, but it will search executable file directory by default. And CMake will ignore BUILD_RPATH property on Windows safely.
- BUILD_RPATH doesn't affect files after `cmake --install`, but we doesn't declare install rule now (And we use RUNTIME_OUTPUT_DIRECTORY so we actually use non-install-style). If we need install-style in future, we can set INSTALL_RPATH_USE_LINK_PATH property so CMake will ensure it will work.

### DIFF
--- a/src/dlsym_interface.c
+++ b/src/dlsym_interface.c
@@ -12,13 +12,10 @@
 
 #if defined(__MINGW32__)
 #define INTERFACELIB_POSTFIX "dll"
-#define DIRSEP "\\"
 #elif defined(__APPLE__)
 #define INTERFACELIB_POSTFIX "dylib"
-#define DIRSEP "/"
 #else
 #define INTERFACELIB_POSTFIX "so"
-#define DIRSEP "/"
 #endif
 
 static struct applet_entry applet_apdu = {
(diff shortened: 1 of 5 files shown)


septs (approved):

src/dlsym_interface.c
```diff
@@ -53,12 +53,12 @@ static void dlsym_interfaces_get_path(void)
...
     if (!(libapduinterface_path = getenv("APDU_INTERFACE")))
     {
-        libapduinterface_path = "." DIRSEP "libapduinterface_pcsc." INTERFACELIB_POSTFIX;
+        libapduinterface_path = "libapduinterface_pcsc." INTERFACELIB_POSTFIX;
     }
 
     if (!(libhttpinterface_path = getenv("HTTP_INTERFACE")))
     {
-        libhttpinterface_path = "." DIRSEP "libhttpinterface_curl." INTERFACELIB_POSTFIX;
+        libhttpinterface_path = "libhttpinterface_curl." INTERFACELIB_POSTFIX;
```
damonto, line 61: Did you test this change on Windows? I remember it might not be working on my Windows machine

damonto: And you should remove the DIRSEP too

CoelacanthusHex: Done. Remove upstream strange LOAD_WITH_ALTERED_SEARCH_PATH and avoid UB.
https://github.com/dlfcn-win32/dlfcn-win32/issues/104

septs (approved):

CoelacanthusHex:
Tested on Windows and Arch Linux, and it works well.

PR #44, opened 2024-03-06
Add Dockerfile for lpac

DenrianWeiss:
Please remember adding port mapping for 1888


BeRealQueally:
suggest to publish by gist

PR #70, opened 2024-04-26
Add QMI-over-QRTR APDU backend for running on Qualcomm SoCs

### LINKED ISSUE #41: Qualcomm QMI/QRTR backend?
Hi,

In the last week I've got lpac to work on two different Qualcomm-based smartphones running that both have an eSIM built-in. Software stack is upstream Linux kernel plus postmarketOS user space.

So far I've written a (not too horrible) wrapper around lpac that uses the stdio interface and handles/forwards the request using qmicli from libqmi. More meant as a proof of concept, I think building a proper QMI backend for lpac would be a good idea - given there's already an AT and PCSC backend for it.

I haven't looked much into it but generally my idea would be to use (glib-based) libqmi (with my patches https://gitlab.freedesktop.org/mobile-broadband/libqmi/-/merge_requests/376) and then the C API of libqmi to do the different operations.

Are there any major concerns I should be aware of?

Alternatively it might be possible to include `.qmi`-based description of the messages and generate C bindings with https://github.com/andersson/qmic and probably also using libqrtr from https://github.com/andersson/qrtr but I'm pretty sure this would be quite a bit more work compared to using libqmi.

z3ntu:
On modern smartphones the communication happens using the QMI protocol via the QRTR transport to communicate with the modem.

Add support for such devices by wrapping libqrtr-glib and libqmi-glib and using those APIs to provide an APDU backend for lpac.

Fixes #41 

---

~~Depends on this libqmi MR which hasn't been merged yet: https://gitlab.freedesktop.org/mobile-broadband/libqmi/-/merge_requests/376~~ merged!


driver/apdu/qmi_qrtr.c
```diff
@@ -0,0 +1,241 @@
...
+#include <euicc/interface.h>
+#include <signal.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <stdbool.h>
+#include <libqrtr-glib.h>
+#include "qmi_qrtr_helpers.h"
+
+#define UIM_SLOT 2 // TODO: Make this runtime-configurable
```
z3ntu, line 16: Do you have a suggestion how to handle the dual-SIM situations, where one wants to set the slot to 1 or 2 (or in theory slot 3 and higher could also exist)?

One idea is adding an environment variable UIM_SLOT=1 to lpac invocation but this wouldn't be nice in my opinion.

AndySchroder:
I have a Quectel EM120R-GL (https://www.quectel.com/wp-content/uploads/2021/05/Quectel_EM120R-GLEM160R-GL_Hardware_Design_V1.1.pdf), is that device I could test this with? It supports eSIM, and I think it has a Qualcomm chip. It says it supports QMI and QRTR.

AndySchroder:
If you have multiple modems, how does it know which one to use?

AndySchroder:
Explain to me the purpose of this feature exactly. We already have ports like `/dev/wwan0at1` and `/dev/ttyUSB3` for running AT commands on the modem. Can we not just use those ports with `at` APDU backend?

z3ntu:
The modems on Qualcomm smartphones don't do AT commands, you communicate with them through QMI.

But for curiosity, does anything show up with `qrtr-lookup` (from https://github.com/linux-msm/qrtr) with your Quectel modem?

AndySchroder:
> But for curiosity, does anything show up with `qrtr-lookup` (from https://github.com/linux-msm/qrtr) with your Quectel modem?

Not sure how to get that installed. There aren't any build instructions. I'm running on ubuntu 22.04.

z3ntu:
@AndySchroder `git clone https://github.com/linux-msm/qrtr && cd qrtr && meson setup builddir && meson compile -C builddir && ./builddir/src/qrtr-lookup` should work, but yeah if it's an external modem then this PR is probably not applicable.

What I know is that some external modems use MBIM so you can check https://github.com/z3ntu/lpac-libqmi-wrapper/issues/1 or even better https://github.com/stich86/lpac-libmbim-wrapper but I don't know any more about eSIM on external modems than that :)

AndySchroder:
```
$ ./builddir/src/qrtr-lookup
ERROR qrtr-lookup: sock(AF_QIPCRTR): Address family not supported by protocol
$
```

AndySchroder:
The data sheet for the EM120R-GL says `QMI/MBIM`. Also, `mmcli` reports `drivers: option, cdc_mbim` and `ports: cdc-wdm0 (mbim), ttyUSB0 (qcdm)`.

estkme:
Just a hearbeat.

We're still waiting for the upstream library to merge your patch, aren't we?

z3ntu:
Yes, I think it would be good if that happens first - before merging this one at least. If you have any comments already on this PR let me know.

z3ntu:
The libqmi MR was just merged so apart from this not being in a release yet (the backend is not built by default anyways), all good now :)

bam80:
@z3ntu 
Could you solve the conflicts?
I just wish we keep this PR ready :)

driver/apdu/qmi_qrtr.c
z3ntu: @estkme Thanks for merging!

But do you have any suggestion how to handle this part? I'm actually thinking now environment variable might be okay since also the AT backend just does `getenv("AT_DEVICE")` for example. 

I'd add a `getenv("UIM_SLOT")` call to this backend then to replace this hardcoded slot 2.

estkme: I think `getenv("UIM_SLOT")`  is okay for now, later our team may have a guide to how to name an ENV var ---- that will be another pr.

z3ntu: Done: https://github.com/estkme-group/lpac/pull/97

bam80:
As this PR introduces quite an important feature, could we make a release maybe?

PR #71, opened 2024-05-03
refactor: profile discovery

septs:


src/utils.c
```diff
@@ -0,0 +1,29 @@
...
+    int count = 0;
+    int allowed;
+    for (unsigned long i = strlen(name); i > 0; i--)
+    {
+        if (name[i] == '.')
+        {
+            count = 0;
+            continue;
+        }
+        allowed = (name[i] >= '0' && name[i] <= '9') ||
```
CoelacanthusHex, line 15: According to RFC1035, if doesn't consider IDN, a domain allow letter, digit and dash. You check function seems not allow dash.

PR #76, opened 2024-05-06
chore: improve build script

### LINKED ISSUE #75: issues with installation instructions
Tried following your installation instructions starting here: https://github.com/estkme-group/lpac/blob/ee4b208186a9a53f918db7391616905d5a83a265/README.md?plain=1#L38 .

So far, I've noticed a few issues:
1. I needed to run `sudo apt update` first.
2. When running `./scripts/build.sh make` I get `./scripts/build.sh: line 53: zip: command not found`. The fix was to run `sudo apt install zip`.
3. No `output` directory exists after compiling. Instead there is a file `lpac-linux-aarch64.zip` that I then need to unzip somewhere else.

septs:
resolve #75

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -50,7 +50,7 @@ case "${1:-}" in
 make)
     cmake "$WORKSPACE"
     make -j
-    zip -r -j "$WORKSPACE/lpac-$KERNEL-$MATCHINE.zip" output/*
+    [ -f "$(which zip)" ] && zip -r -j "$WORKSPACE/lpac-$KERNEL-$MATCHINE.zip" output/*
     ;;
 debian)
     cmake "$WORKSPACE" -DCPACK_GENERATOR=DEB
@@ -62,7 +62,7 @@ mingw)
     make -j
     CURL="$(download "$MINGW_CURL_WIN64_BLOB")"
     cp "$CURL"/curl-*-mingw/bin/libcurl-x64.dll output/libcurl.dll
-    zip -r -j "$WORKSPACE/lpac-windows-x86_64-mingw.zip" output/*
+    [ -f "$(which zip)" ] && zip -r -j "$WORKSPACE/lpac-windows-x86_64-mingw.zip" output/*
     ;;
 woa-mingw)
     TOOLCHAIN="$(download "$MINGW32_TOOLCHAIN_BLOB")"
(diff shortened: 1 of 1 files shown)


scripts/build.sh
```diff
@@ -62,27 +62,31 @@ mingw)
...
     ;;
 *)
     echo "Usage: $0 {make,debian,mingw,woa-mingw,woa-zig}"
     exit 1
     ;;
 esac
 
+rm -rf "$WORKSPACE/build"
+mkdir "$WORKSPACE/build"
+cp -r "$BUILD"/* "$WORKSPACE/build"
```
AndySchroder, line 90: why don't you move instead of copy if you are just going to delete below?

AndySchroder:
Does this address the need for `sudo apt update` and `sudo apt install zip`?

AndySchroder:
The need for `sudo apt update` is a separate issue

AndySchroder:
But if zip is installed, then you have to zip the output and then the user will have to re-unzip it to actually use the compiled file? Seems a little weird way to do it to me.

septs:
https://github.com/estkme-group/lpac/blob/ee4b208186a9a53f918db7391616905d5a83a265/.github/workflows/build.yaml#L27-L33

septs:
> The need for `sudo apt update` is a separate issue

what does it mean? hint install package before need update apt database?

AndySchroder:
> https://github.com/estkme-group/lpac/blob/ee4b208186a9a53f918db7391616905d5a83a265/.github/workflows/build.yaml#L27-L33

Not sure what you are referencing here, but it seems to me like zipping or not zipping the output should be a build flag and not just doing it if a zip program is installed or not installed.

AndySchroder:
> > The need for `sudo apt update` is a separate issue
> 
> what does it mean? hint install package before need update apt database?

This is a command you normally run before running `sudo apt install`. It refreshes the repository database so that when you try to `sudo apt install` packages, it can actually find them and find the latest version. Usually you have to run it once a day. In my case I haven't run it in several days so `sudo apt install` errored out. Since you never know when the last time someone ran `sudo apt update`, installation instructions should always say to do it every time. It only wastes a few seconds redoing it and no other harm done.

scripts/build.sh
```diff
@@ -62,27 +62,36 @@ mingw)
...
     exit 1
     ;;
 esac
 
+rm -rf "$WORKSPACE/build"
+mkdir "$WORKSPACE/build"
+if [ -d "$BUILD/output" ]; then
+    cp -r "$BUILD/output" "$WORKSPACE/build"
+fi
+if [ -n "$(find "$BUILD" -name 'lpac-*.deb')" ]; then
```
AndySchroder, line 93: nice to know that a .deb package can be created, but how do I even create that? I don't see it listed anywhere in the installation instructions.

AndySchroder:
> > > The need for `sudo apt update` is a separate issue
> > 
> > 
> > what does it mean? hint install package before need update apt database?
> 
> This is a command you normally run before running `sudo apt install`. It refreshes the repository database so that when you try to `sudo apt install` packages, it can actually find them and find the latest version. Usually you have to run it once a day. In my case I haven't run it in several days so `sudo apt install` errored out. Since you never know when the last time someone ran `sudo apt update`, installation instructions should always say to do it every time. It only wastes a few seconds redoing it and no other harm done.

For example, you show `sudo apt update` here https://github.com/estkme-group/lpac/blob/ee4b208186a9a53f918db7391616905d5a83a265/scripts/setup-debian.sh#L3 . However, I'm not sure what that script is actually for.

PR #88, opened 2024-05-09
feat: cancel session in download failed

septs:


src/applet/profile/download.c
```diff
@@ -130,6 +137,8 @@ static int applet_main(int argc, char **argv)
         goto err;
     }
 
+    signal(SIGINT, on_cancel_session);
```
CoelacanthusHex, line 140: Use `sigaction(2)` instaed of `signal(2)` if you use a custom handler function rather thanSIG_DFL or SIG_IGN because it's nonportable, see also section VERSION in https://man7.org/linux/man-pages/man2/signal.2.html.
`signal(2)` also has other issue like https://glandium.org/blog/?p=2979.

PR #104, opened 2024-05-28
Add a prefix to each environment variable.

damonto:


driver/apdu/gbinder_hidl.c
```diff
@@ -27,19 +27,22 @@
 
 static int lastChannelId = -1;
 
-struct radio_response_info {
+struct radio_response_info
+{
```
CoelacanthusHex, line 31: why change code style?

damonto: My editor formatted it automatically.

CoelacanthusHex: Looks like we need a .clang-format config.

PR #121, opened 2024-06-16
chore: disable fail-fast and better artifacts naming in action

BeRealQueally:


.github/workflows/build.yaml
```diff
@@ -26,15 +27,21 @@ jobs:
...
+
       - name: Setup Toolchain
         run: ./scripts/setup-debian.sh
         if: runner.os == 'Linux'
+
       - name: Build for ${{ matrix.build.name }}
         run: ./scripts/build-ci.sh ${{ matrix.build.variant }}
+
+      - name: Get short SHA
+        run: echo "SHA7=$(echo ${GITHUB_SHA} | cut -c1-7)" >> $GITHUB_ENV
```
septs, line 39: why not `git describe --always`?

septs: more see https://www.git-scm.com/docs/git-describe

PR #123, opened 2024-06-21
refactor: bash scripts

septs:


.github/workflows/build.yaml
```diff
@@ -29,7 +29,7 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Setup Toolchain
-        run: ./scripts/setup-debian.sh
+        run: sudo ./scripts/setup-debian.sh ${{ matrix.build.variant }}
```
CoelacanthusHex, line 32: Why？

scripts/build-ci.sh
```diff
@@ -1,66 +1,25 @@
...
-        ;;
-    *)
-        echo "$SAVED_PATH"
-        ;;
-    esac
-}
+source "$SCRIPT_DIR/functions.sh"
 
-set -x
+set -xeuo pipefail
```
CoelacanthusHex, line 6: `set -x` will pollute output, use it only when debugging.

.github/workflows/build.yaml
septs: the script internal use `sudo apt-get`, this is move `sudo` to the script external.

scripts/build-ci.sh
septs: i want the user to know which commands were run

.github/workflows/build.yaml
CoelacanthusHex: AFAIK, `sudo` is useless in  GitHub Actions, it's enough to just run `apt-get`.

scripts/build-ci.sh
CoelacanthusHex: The user should read the script, instead of DoS the brain of maintainers.

.github/workflows/build.yaml
septs: OK, i try it

scripts/build-ci.sh
septs: OK

.github/workflows/build.yaml
septs: CleanShot 2024-06-21 at 13 49 22@2x

PR #129, opened 2024-07-03
Grow buffer for retreiving AT modem replies.

steely-glint:
The replies to some commands (eg profile list) can exceed 1k - in which case the parser fails to see the line ending and no response is returned.

(tested on Quectel RM520 module)


damonto:
#140

PR #131, opened 2024-07-03
 driver: apdu: add QMI backend

robimarko:
This patch series adds support for using QMI interface directly.

Previously QMI over QRTR support was added, however that is only present in
modern Qualcomm modems and only when running in PCIe mode and its quite
common to still only use even the latest modems via USB.

In that case, they are all still controllable via QMI but it is directly
exposed as a character device in Linux so we can reuse most of the code
from QMI over QRTR support but drop the support for libqrtr to talk to
the modems.

We require the QMI device path to be passed via QMI_DEVICE env variable
for the backend to operate.

Most commits in the series are to refactor the current QMI code in order to avoid duplicating core,
or to get rid of using global variables for the common code.


robimarko:
Fixed up the build warning

septs:
if you are ready, plz click "Ready for review" convert the PR state

PR #133, opened 2024-07-13
chore(driver): Improve AT command APDU driver

damonto:


driver/apdu/at.c
```diff
@@ -50,32 +188,24 @@ static int apdu_interface_connect(struct euicc_ctx *ctx)
 
     if (!(device = getenv("AT_DEVICE")))
     {
-        device = "/dev/ttyUSB0";
+        device = "/dev/ttyUSB2";
```
septs, line 191: why modify default AT device value?

driver/CMakeLists.txt
```diff
@@ -2,9 +2,13 @@ include(CMakeDependentOption)
...
 
 option(LPAC_WITH_APDU_PCSC "Build APDU PCSC Backend (requires PCSC libraries)" ON)
-option(LPAC_WITH_APDU_AT "Build APDU AT Backend" ON)
 option(LPAC_WITH_APDU_GBINDER "Build APDU Gbinder backend for libhybris devices (requires gbinder headers)" OFF)
 option(LPAC_WITH_APDU_QMI_QRTR "Build QMI-over-QRTR backend for Qualcomm devices (requires libqrtr and libqmi headers)" OFF)
+if (Linux)
+    option(LPAC_WITH_APDU_AT "Build APDU AT Backend" ON)
+else()
+    option(LPAC_WITH_APDU_AT "Build APDU AT Backend" OFF)
+endif()
```
septs, line 11:
```suggestion
option(LPAC_WITH_APDU_AT "Build APDU AT Backend" OFF)

if (Linux)
    set(LPAC_WITH_APDU_AT ON)
endif()
```

disabled by default, but if is linux then enabled by default

see https://cmake.org/cmake/help/latest/command/option.html
see https://cmake.org/cmake/help/latest/command/set.html

CoelacanthusHex, line 11:
```suggestion
option(LPAC_WITH_APDU_AT "Build APDU AT Backend" LINUX)
```

driver/apdu/at.c
```diff
@@ -9,35 +9,170 @@
...
+static int fd = -1;
 static int logic_channel = 0;
 
+// Function to open AT device
+static int open_at_device(const char *device)
+{
+    fd = open(device, O_RDWR | O_NOCTTY);
+    if (fd == -1)
+    {
+        fprintf(stderr, "Failed to open device: %s\n", device);
```
CoelacanthusHex, line 28:
```suggestion
        fprintf(stderr, "Failed to open device: %s: %s\n", device, strerror(errno));
```
It may be better to print the exact reason at the same time. Also for error handle code below.

CoelacanthusHex (commented):
And note the `open()`, `close()`, `tcflush()`, `tcgetattr()` all are thing of POSIX. There is no such thing in Windows. `flock()` is even not a part of POSIX, although it's a part of both Linux and BSD.

Issue #138, opened 2024-08-01
AT command APDU driver violates 3GPP TS 27.007 and will not work on e.g. MC7455


laf0rge:
The AT command APDU driver seems to misunderstand how 3GPP TS 27.007 works.  It assumes that the unsoliciated `+CCHO` response will return the logical channel number, and will then by itself patch that number into the CLA byte.

However, 3GPP TS 27.007 works completely differently:
* the logical channel number is only visible between Modem (MT) and card (eUICC)
* the `+CCHO` unsolicited response returns a "handle", which is of local significance only between the modem and the application issuing the AT commands
* the application must sent CLA bytes where the logical channel bits are zero; the MT (Modem) will patch the logical number into it

For example, see the following exchange with a Sierra MC7455 modem:
```
AT_DEBUG: AT+CCHO="A0000005591010FFFFFFFF8900000100"
AT_DEBUG: +CCHO: 000000000385a7cf
```

clearly the 0x385a7cf is *not* a logical channel number, but a handle that has no significance other than it must be used for all other subsequent operations like `+CCHC` or `+CGLA`

Spec reference: **3GPP TS 27.007 Rel 17.8.0 Section 8.45** which specificly states what I wrote
above:
> In this case it will be up to the MT to manage the logical channel part of the APDU CLASS byte and to ensure that the chosen logical channel is relevant to the <sessionid> indicated in the AT command.

To me it looks like the AT command driver was not written against what TS 27.007 specifies, but maybe against one specific modem implementation that may itself return the logical channel in the `AT+CCHO`.  However, that is pure coincidence and cannot be assumed in general

urielka:
@laf0rge look at my patch:
https://github.com/estkme-group/lpac/pull/139

Not the cleanest but at least it is working and might be useful for other poor folks that run into this issue

septs:
> Defined values
>
> <sessionid>: integer type; a session Id to be used in order to target a specific application on the smart card (e.g. (U)SIM, WIM, ISIM) using logical channels mechanism 

from [ETSI TS 127 007 V15.2.0 (2018-06)](https://www.etsi.org/deliver/etsi_ts/127000_127099/127007/15.02.00_60/ts_127007v150200p.pdf#page=150) (Page 150 of 368)

septs:
i think this is a violation of the baseband specification

PR #139, opened 2024-08-01
Working AT driver for MC7455 and EM7590, not tested with anything else

urielka:
Seems like with sierra wireless there are 3 issues with the existing AT code:
1. They work with logic channels that aren't integers, they use a big hex number
```
AT_DEBUG: AT+CCHO="A0000005591010FFFFFFFF8900000100"
AT_DEBUG: +CCHO: 00000000038aa32f

```
2. The code assumes that the number returned from CCHO is the logic channel number for the eUICC and should be added to the CLA base (0x80) which results in cases sending 0x8F which seems to work with *some* eUICC.
see: https://github.com/estkme-group/lpac/issues/138 for explanation
3. The code checks for AT+CGLA=? running OK as a way to detect if AT+CGLA= is supported but it doesn't mean that. in EM7590 AT+CGLA= is supported and AT+CGLA=? isn't

The following code was only tested with MC7455 and EM7590. tested with esim.me sim and sysmocom (https://shop.sysmocom.de/sysmoEUICC1-eUICC-for-consumer-eSIM-RSP/sysmoEUICC1-C2G)

### DIFF
--- a/driver/apdu/at.c
+++ b/driver/apdu/at.c
@@ -11,7 +11,8 @@
 
 #define AT_BUFFER_SIZE 20480
 static FILE *fuart;
-static int logic_channel = 0;
+
+static long logic_channel = 0;
 static char *buffer;
 
 static int at_expect(char **response, const char *expected)
@@ -74,13 +75,15 @@ static int apdu_interface_connect(struct euicc_ctx *ctx)
         fprintf(stderr, "Device missing AT+CCHC support\n");
         return -1;
     }
+    /* AT+CGLA=? might not be available but AT+CGLA= is
+       Sierra wireless MC7455 supports both but EM7590 only the latter
     fprintf(fuart, "AT+CGLA=?\r\n");
     if (at_expect(NULL, NULL))
     {
         fprintf(stderr, "Device missing AT+CGLA support\n");
         return -1;
     }
-
+    */
     return 0;
 }
 
@@ -106,7 +109,7 @@ static int apdu_interface_transmit(struct euicc_ctx *ctx, uint8_t **rx, uint32_t
         return -1;
     }
 
-    fprintf(fuart, "AT+CGLA=%d,%u,\"", logic_channel, tx_len * 2);
+    fprintf(fuart, "AT+CGLA=%lx,%u,\"", logic_channel, tx_len * 2);
     for (uint32_t i = 0; i < tx_len; i++)
     {
         fprintf(fuart, "%02X", (uint8_t)(tx[i] & 0xFF));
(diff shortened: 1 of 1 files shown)


urielka:
Reverted long usage in logic_channel_open interface as we always should return 0.

Hope now it builds on all platforms

driver/apdu/at.c
```diff
@@ -187,9 +192,13 @@ static int apdu_interface_logic_channel_open(struct euicc_ctx *ctx, const uint8_
     {
         return -1;
     }
-    logic_channel = atoi(response);
 
-    return logic_channel;
+    logic_channel = strtol(response, NULL, 16);
```
CoelacanthusHex, line 196:
```suggestion
    logic_channel = strtoull(response, NULL, 16);
```
Same as above.

driver/apdu/at.c
```diff
@@ -198,7 +207,7 @@ static void apdu_interface_logic_channel_close(struct euicc_ctx *ctx, uint8_t ch
     {
         return;
     }
-    fprintf(fuart, "AT+CCHC=%d\r\n", logic_channel);
+    fprintf(fuart, "AT+CCHC=%lx\r\n", logic_channel);
```
CoelacanthusHex, line 210:
```suggestion
    fprintf(fuart, "AT+CCHC=%" PRIu "\r\n", logic_channel);
```
Same as above.

driver/apdu/at.c
```diff
@@ -74,13 +75,15 @@ static int apdu_interface_connect(struct euicc_ctx *ctx)
...
+    /* AT+CGLA=? might not be available but AT+CGLA= is
+       Sierra wireless MC7455 supports both but EM7590 only the latter
     fprintf(fuart, "AT+CGLA=?\r\n");
     if (at_expect(NULL, NULL))
     {
         fprintf(stderr, "Device missing AT+CGLA support\n");
         return -1;
     }
-
+    */
```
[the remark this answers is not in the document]
urielka: I don't think it is that important to be honest, it will just fail after the first command, checking for AT+CGLA= requires to actually do AT+CCHO and then AT+CGLA so what are we gaining here.

Maybe change it to a warning? and again the issue is the lack of AT+CGLA= and not AT+CGLA=? where in AT commands those are separated commands so just like in the case of sierra wireless it might not be implemented

septs (changes requested):
We cannot assume that all baseband modules are designed based on the MC7455 / EM7590 AT command set.

CoelacanthusHex:
It's better to split this PR into three PRs:

1. int can't deal with the return value of `AT+CCHO` on some modems, use int64_t instead.
2. The return value of `AT+CCHO` shouldn't be treated as the logic channel, address https://github.com/estkme-group/lpac/issues/138
3. `AT+CGLA=?` works don't mean `AT+CGLA=` is supported.
4. 
So we could review them case by case.

septs:
long unprocessed review

PR #141, opened 2024-08-03
feat: support for preview mode

damonto:


src/applet/profile/download.c
```diff
@@ -9,24 +9,26 @@
...
     int fret;
-
     int opt;
 
     char *smdp = NULL;
     char *matchingId = NULL;
     char *imei = NULL;
     char *confirmation_code = NULL;
     char *activation_code = NULL;
+    int is_preview_mode = -1;
```
CoelacanthusHex, line 27: It's better to use bool since we use C99. Incluse `stdbool.h` and use `bool` type, or use ugly `_Bool` directly.

PR #149, opened 2024-08-29
driver: add uqmi backend

blocktrron:
This commit adds a backend to interface with an eUICC connected to a compatible modem using the uqmi application.

This allows OpenWrt to manage an eUICC including the download and installation of SIM profiles.

This was previously not possible with most modems, as the APDU operations were aborted due to exceeding the timeout imposed on the AT interface by the modem firmware.

Tested-on: Quectel EC25 / Quectel EP06 / Quectel RG520N

### DIFF
--- a/driver/CMakeLists.txt
+++ b/driver/CMakeLists.txt
@@ -6,6 +6,7 @@ option(LPAC_WITH_APDU_AT "Build APDU AT Backend" Linux)
 option(LPAC_WITH_APDU_GBINDER "Build APDU Gbinder backend for libhybris devices (requires gbinder headers)" OFF)
 option(LPAC_WITH_APDU_QMI "Build QMI backend for Qualcomm devices (requires libqmi)" OFF)
 option(LPAC_WITH_APDU_QMI_QRTR "Build QMI-over-QRTR backend for Qualcomm devices (requires libqrtr and libqmi headers)" OFF)
+option(LPAC_WITH_APDU_UQMI "Build uQMI backend for Qualcomm devices (requires uqmi)" OFF)
 
 option(LPAC_WITH_HTTP_CURL "Build HTTP Curl interface" ON)
 
@@ -47,8 +48,10 @@ if(LPAC_WITH_APDU_AT)
     target_sources(euicc-drivers PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/apdu/at.c)
 endif()
 
-set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -DLPAC_WITH_APDU_UQMI")
-target_sources(euicc-drivers PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/apdu/uqmi.c)
+if(LPAC_WITH_APDU_UQMI)
+    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -DLPAC_WITH_APDU_UQMI")
+    target_sources(euicc-drivers PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/apdu/uqmi.c)
+endif()
 
 if(LPAC_WITH_APDU_GBINDER)
     set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -DLPAC_WITH_APDU_GBINDER")
--- a/driver/CMakeLists.txt
+++ b/driver/CMakeLists.txt
@@ -47,6 +47,9 @@ if(LPAC_WITH_APDU_AT)
     target_sources(euicc-drivers PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/apdu/at.c)
 endif()
 
+set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -DLPAC_WITH_APDU_UQMI")
+target_sources(euicc-drivers PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/apdu/uqmi.c)
+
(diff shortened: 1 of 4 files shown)


septs (changes requested):
this is safe from command injection attacks?

see https://owasp.org/www-community/attacks/Command_Injection

blocktrron:
@septs Where do you see the potential for Command injection? APDU data is generated by the existing lpac infrastructure and all binary data potentially landing in the system shell is formatted only in hex-chars here:

https://github.com/estkme-group/lpac/pull/149/files#diff-ec954341e60cb55be4a97cf02c58237678cc535c127029fe5a6c95fd228c34f2R125

The buffer is fixed-size 2048 byte long, which fits the maximum tansport size. However, it currently does not check if tx_len fits in there, I will fix this.

To give a it of background: `uqmi` is a standalone tool for modem-control on OpenWrt, so it sadly does not have a linkable library we could use and make the detour over the shell. This is set to change, as uqmi is being rewritten to provide an rpc interface. For the time being however, uqmi is the best way to interact with eUICCs in modems with small flash constraints like the ones targeted by OpenWrt.

Still looking forward for your feedback!

driver/apdu/uqmi.c
```diff
@@ -0,0 +1,273 @@
...
+    if (debug)
+        printf("UQMI_DEBUG_TX: %s\n", command);
+
+    if (snprintf(final_command, sizeof(final_command), "uqmi -s -d %s %s", devpath, command) >= sizeof(final_command))
+    {
+        fprintf(stderr, "Command too long\n");
+        return -1;
+    }
+
+    fp = popen(final_command, "r");
```
CoelacanthusHex, line 40: We should avoid `system(3)` or `popen(3)` because they pass commandline to shell and then shell will do some extra things like wordsplit, glob expansion. It doesn't execute as what we write.

driver/apdu/uqmi.c
```diff
@@ -0,0 +1,273 @@
...
+    for (int i = 0; i < tx_len; i++)
+    {
+        snprintf(buf + strlen(buf), sizeof(buf) - strlen(buf), "%02X", (uint8_t)(tx[i] & 0xFF));
+    }
+
+    snprintf(cmd, sizeof(cmd), "--set-client-id uim,%d --keep-client-id uim --uim-slot 1 --uim-channel-id %d --uim-apdu-send \"%s\"",
+             client_id, logic_channel, buf);
+    uqmi_execute_command(cmd, buf, sizeof(buf));
+
+    json = strstr(buf, "{");
```
CoelacanthusHex, line 132: I think it's unnecessary since it has been covered by `cJSON_Parse()`.

driver/apdu/uqmi.c
```diff
@@ -0,0 +1,273 @@
...
+
+    if (!logic_channel)
+    {
+        return -1;
+    }
+
+    for (int i = 0; i < tx_len; i++)
+    {
+        snprintf(buf + strlen(buf), sizeof(buf) - strlen(buf), "%02X", (uint8_t)(tx[i] & 0xFF));
+    }
```
septs, line 126: builtin bin to hex string function, plz use `euicc_hexutil_bin2hex`

https://github.com/estkme-group/lpac/blob/9e0612b68dec3f02e8fac8997d8dd653c369f42b/euicc/hexutil.h#L4-L8

driver/apdu/uqmi.c
septs: plz use `spawn(3)`

see https://man7.org/linux/man-pages/man3/posix_spawn.3.html

driver/CMakeLists.txt
```diff
@@ -47,6 +47,9 @@ if(LPAC_WITH_APDU_AT)
     target_sources(euicc-drivers PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/apdu/at.c)
 endif()
 
+set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -DLPAC_WITH_APDU_UQMI")
+target_sources(euicc-drivers PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/apdu/uqmi.c)
+
```
septs, line 52:
```suggestion
if(LPAC_WITH_APDU_UQMI)
  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -DLPAC_WITH_APDU_UQMI")
  target_sources(euicc-drivers PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/apdu/uqmi.c)
endif()
```

uQMI is optional and not available on all platforms

z3ntu:
Is there any difference to https://github.com/estkme-group/lpac/pull/131 apart from using uqmi instead of libqmi? The functionality should be the same, right?

septs:
> Is there any difference to #131 apart from using uqmi instead of libqmi? The functionality should be the same, right?

uQMI is OpenWrt specific

septs:
long unprocessed review

blocktrron:
@septs Sorry for taking this long. I still ave this on my backlog and will notify you once i have an updated branch ready. Thank you for your review and work!

septs:
i think safety code style like:

```c
char *client_id = ...;
char *slot = ...;
char *channel_id = ...;

char *args = {
   "--set-client-id", client_id,
   "--keep-client-id", "uim",
   "--uim-slot", slot,
   "--uim-channel-id", channel_id,
   "--uim-apdu-send"
};
uqmi_execute_command(args, buffer, sizeof(buffer));
```

PR #150, opened 2024-09-01
Update LINUX-DIST.md with debianization link

danielg4:
Apparently, someone known as erebion.eu made a competing debianization at the exact same time, but this is what I did

### DIFF
--- a/docs/LINUX-DIST.md
+++ b/docs/LINUX-DIST.md
@@ -53,7 +53,7 @@ see <https://github.com/nix-community/nur-combined/blob/master/repos/linyinfeng/
 
 ## Ubuntu and Debian/Devuan
 
-> Minimum available release: 14.04 Trusty Tahr
+> Minimum available release: Ubuntu [14.04 Trusty Tahr](https://releases.ubuntu.com/14.04/)
 > (Published on 2024-09-01)
 
-see <https://launchpad.net/~daniel-gimpelevich/+archive/ubuntu/ssl/+sourcepub/16421117/+listing-archive-extra>
+see <https://launchpad.net/~daniel-gimpelevich/+archive/ubuntu/ssl/+sourcepub/16426799/+listing-archive-extra>
--- a/docs/LINUX-DIST.md
+++ b/docs/LINUX-DIST.md
@@ -50,3 +50,10 @@ nix-env -i lpac
 ```
 
 see <https://github.com/nix-community/nur-combined/blob/master/repos/linyinfeng/pkgs/lpac/default.nix>
+
+## Ubuntu and Debian/Devuan
+
+> Minimum available release: 14.04 Trusty Tahr
+> (Published on 2024-09-01)
+
+see <https://launchpad.net/~daniel-gimpelevich/+archive/ubuntu/ssl/+sourcepub/16421117/+listing-archive-extra>


erebion:
For safety, I'd recommend not linking to a random Launchpad profile.

danielg4:
Indeed, which is why the link was not to the profile but to the packaging itself. It's as preliminary as can be, and once it's in its final form, the link should be replaced.

docs/LINUX-DIST.md
```diff
@@ -50,3 +50,10 @@ nix-env -i lpac
 ```
 
 see <https://github.com/nix-community/nur-combined/blob/master/repos/linyinfeng/pkgs/lpac/default.nix>
+
+## Ubuntu and Debian/Devuan
+
+> Minimum available release: 14.04 Trusty Tahr
```
septs, line 56:
```suggestion
> Minimum available release: [14.04 Trusty Tahr](https://releases.ubuntu.com/14.04/)
```

danielg4: LGTM. The reason I thought it OK to leave out was that it's left out in the OpenWrt section.

septs: Snapshot not is General available

septs: Debian codename and Ubuntu codename is inconsistent

I think should separate

erebion:
Meanwhile, I am working on it at https://salsa.debian.org/erebion/lpac , will probably package it without qmi_qrtr for now, as libqmi is not yet available in Debian in the required version.

docs/LINUX-DIST.md
danielg4: > OpenWrt Snapshot are not generally available

OpenWrt snapshot is the counterpart of Debian Sid. Debian Sid is very available.

danielg4: > Debian codename and Ubuntu codename is inconsistent
> 
> I think should separate

The release schedules were always different. Trusty Tahr is from more than 10 years ago, and everyone will be using something far more recent. The package I made should work the same on every version of both since then.

danielg4:
> Meanwhile, I am working on it at https://salsa.debian.org/erebion/lpac , will probably package it without qmi_qrtr for now, as libqmi is not yet available in Debian in the required version.

Without the qmi driver, it would probably be functionally identical to the package I made. Adding that driver would be the major benefit of making it an official part of Debian.

erebion:
Just qmi_qrtr would be missing for now. :)

The other drivers are currently compiled in.

septs (approved):
wait for @CoelacanthusHex

CoelacanthusHex:
IMO, it's better to wait for official debian packaging of @erebion.

danielg4:
> Just qmi_qrtr would be missing for now. :)
> 
> The other drivers are currently compiled in.

Whoops! The AT driver was evidently not compiled into my package by mistake. That and other packaging oversights on my part corrected:
https://launchpad.net/~daniel-gimpelevich/+archive/ubuntu/ssl/+sourcepub/16426799/+listing-archive-extra
Note the slightly different URL.

erebion:
> wait for @CoelacanthusHex

https://salsa.debian.org/erebion/lpac

Already working here and upload will probably happen soon-ish... end of the month perhaps. :)

PR #152, opened 2024-09-07
Export libeuicc all symbols

ShiinaSekiu:


CoelacanthusHex:
Have you tried https://cmake.org/cmake/help/latest/prop_tgt/WINDOWS_EXPORT_ALL_SYMBOLS.html ?

PR #169, opened 2024-11-18
ensure correct cmake script argument

fransklaver:
When running a build in a scenario with multiple paths in CMAKE_MODULE_PATH, we aren't actually running the git-version script. This happens because we are APPENDing our own module path to the existing setting.

Instead of relying on the contents of CMAKE_MODULE_PATH, create a variable that is dedicated to our own location, and use that to specify the git-version script.


CoelacanthusHex (approved):
LGTM

Issue #167, opened 2024-11-17
AT command APDU has trouble communicating with Quectel modem


ecm0708:
There's a problem using the AT command APDU with quectel modems.  It seems to be related to the method of writing to the serial device.

@steely-glint has a fork which solves the problem.  In driver/apdu/at.c, it replaces fprintf(fuart...) with possibly smaller sized write(uart_fd...).    After merging this fork with the current upstream and reducing the maximum segment size ctx->es10x_mss, it works well on a Quectel modem.  I haven't tested on another modem.  Hopefully, this method of communicating through the serial device would also be compatible with other modems.  (Or maybe an environment variable could be used to select the method.)  It looks like the fork on its own does not work since it doesn't have the ability to [reduce ctx->ex10x_mss from upstream](https://github.com/estkme-group/lpac/commit/a5a0516f084936e7e87cf7420fb99283fa3052ef).  

Would it be feasible to merge this upstream?  If needed, I could provide a patch or PR with the merge conflicts resolved.
https://github.com/estkme-group/lpac/compare/main...pipe:lpac-quectel:main#diff-d85497acc8400d3fa4f69b00a762a91dc21f1ef20bdacc9e37cf46218c529287R99

CoelacanthusHex:
> In driver/apdu/at.c, it replaces fprintf(fuart...) with possibly smaller sized write(uart_fd...).

It may be caused by buffered io. Invoking `setbuf(uart_fd, NULL)` to disable buffered I/O after `fopen` should fix the issue.

septs:
this depends on the at modem manufacturer's implementation

ecm0708:
After a long struggle resolving this, it is solved by three changes:

1. https://github.com/CoelacanthusHex/lpac/commit/6281aecda052866ad7293776a277e6ebf38b9f78  
2. Decreasing ctx->es10x_mss  #192
3. Increasing AT_BUFFER_SIZE to about 40k  ([drivers/apdu/at.c](https://github.com/estkme-group/lpac/blob/f9263ad734db50b99b31edad64cb22e50bf3e1ae/driver/apdu/at.c#L12))

What would be the preferred way to implement item 3?  An environment variable (like item 2) or a universal change like item 1?  I don't know the background about how the value for AT_BUFFER_SIZE was selected.

septs:
@CoelacanthusHex `AT_BUFFER_SIZE` needs expose as env vars?

[...]

PR #180, opened 2025-01-03
Fix AT driver

Marekkon5:
The AT driver was very unreliable, sending only partial commands on Quectel devices. This patch seems to fix it for the most part.

### DIFF
--- a/driver/apdu/at.c
+++ b/driver/apdu/at.c
@@ -55,7 +55,7 @@ static int apdu_interface_connect(struct euicc_ctx *ctx)
         device = "/dev/ttyUSB0";
     }
 
-    fuart = fopen(device, "r+");
+    fuart = fopen(device, "rb+");
     if (fuart == NULL)
     {
         fprintf(stderr, "Failed to open device: %s\n", device);


driver/apdu/at.c
```diff
@@ -55,7 +55,7 @@ static int apdu_interface_connect(struct euicc_ctx *ctx)
         device = "/dev/ttyUSB0";
     }
 
-    fuart = fopen(device, "r+");
```
CoelacanthusHex, line 58: It's strange. The "b" flag is a no-op on the POSIX system. It only affects Windows and lets it disable special handling of '\n' and '\x1A'.

CoelacanthusHex (changes requested):
Partially writing is not related to this change completely. That's because `fopen()` produces a buffered IO. The case and solution have been discussed at https://github.com/estkme-group/lpac/issues/167

driver/apdu/at.c
Marekkon5: Although I am unfamiliar with inner workings of the fopen/fread family of functions, this solved the communication issues in my case on Linux. Might've been just luck or some other side effect fixing it. Feel free to close this PR if you don't think this is the real fix.

CoelacanthusHex: The "b" flag is a no-op on the POSIX system because POSIX has no special handling of '\n' and '\x1A' like Windows. It always behaves like Windows with a "b" flag.
https://github.com/bminor/glibc/blob/c2567b18fd77da9d4a270b20232f12616763bf76/libio/fileops.c#L262-L264
Adding `setbuf(fuart, NULL)` after `fopen()` should fix this problem. You can try it and update the PR.

septs: @CoelacanthusHex close the pr?

Marekkon5: After some more testing, on another device neither rb+ or setbuf fixed this PR.

Closing, since solution was caused by some external factor / not reliably reproducible.

PR #183, opened 2025-01-20
Update USAGE.md

neilzone:
Indicate the need to use single quotes when using the `-a` switch for `lpac profile download`, to prevent the shell treating the $ in the activation code string as a variable.

### DIFF
--- a/docs/USAGE.md
+++ b/docs/USAGE.md
@@ -163,7 +163,7 @@ There is no secondary confirmation for deleting a Profile, so please perform it 
 - `-m`: Matching ID, activation code. optional.
 - `-c`: Confirmation Code, optional.
 - `-i`: The IMEI of the device to which Profile is to be downloaded, optional.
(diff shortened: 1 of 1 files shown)


septs:
the is bash or zsh behavior

neilzone:
Yes, that's right - having it in the docs would be a useful prompt when hitting the error messages that I came across today, while trying to use the command as shown in the documents without single quoting it.

drjasonharrison:
@septs would you accept a pull request from @neilzone to add something like:

"Tip: your shell may attempt to interpret the $ in the activation code, or any other special character, typically single quotes will prevent this interpretation."

septs:
i think, this command line utility is suitable for developer, use for example and evaluation

typically used as a C library (e.g: {Open,Easy}EUICC) or as a command line backend (e.g: EasyLPAC)

PR #189, opened 2025-02-09
feat: human readable error code (pcsclite only)

septs:
> For a human readable representation of an error the function pcsc_stringify_error() is declared in pcsclite.h.
>
> This function is not available on Microsoft(R) winscard API and is pcsc-lite specific.

from https://github.com/LudovicRousseau/PCSC/blob/f0f4fa1586f1cc695ad42c88d19dc4f15bfa3c71/src/winscard.c#L68-L70

## Effect

```console
$ ./output/lpac chip
SCardConnect() failed: 80100009 (Unknown reader specified.)
{"type":"lpa","payload":{"code":-1,"message":"euicc_init","data":""}}
```

### DIFF
--- a/driver/apdu/pcsc.c
+++ b/driver/apdu/pcsc.c
@@ -29,6 +29,17 @@ static SCARDCONTEXT pcsc_ctx;
 static SCARDHANDLE pcsc_hCard;
 static LPSTR pcsc_mszReaders;
 
+static char *pcsc_error_text(const int32_t code) {
+    char *formatted = malloc(100);
+    if (!formatted) return NULL;
+#ifdef __MINGW32__
+    snprintf(buf, 8, "%08X",  code);
+#else
+    snprintf(formatted, 100, "%08X (%s)", code, pcsc_stringify_error(code));
+#endif
+    return formatted;
+}
+
 static int pcsc_ctx_open(void)
 {
     int ret;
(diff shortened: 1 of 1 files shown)


driver/apdu/pcsc.c
```diff
@@ -104,13 +112,13 @@ static int pcsc_iter_reader(int (*callback)(int index, const char *reader, void
 static int pcsc_open_hCard_iter(int index, const char *reader, void *userdata)
 {
     int ret;
-    int id;
```
CoelacanthusHex, line 107: It changes nothing on Windows. Windows uses LLP64 data model, it means sizeof(int) == sizeof(long) == 4.

driver/apdu/pcsc.c
```diff
@@ -29,6 +29,14 @@ static SCARDCONTEXT pcsc_ctx;
 static SCARDHANDLE pcsc_hCard;
 static LPSTR pcsc_mszReaders;
 
+static void pcsc_error_text(const char *method, const int32_t code) {
```
CoelacanthusHex, line 32:
```suggestion
static void pcsc_error(const char *method, const int32_t code) {
```
`_text` is confusing.

driver/apdu/pcsc.c
septs: > Clang-Tidy: Narrowing conversion from 'long' to signed type 'int' is implementation-defined

CoelacanthusHex: I mean you should use `long long` and `strtoll()`. `long` should be avoid in cross-platform programming. And it may be better to use `unsigned long long` and `strtoull()` since it's an index.

PR #196, opened 2025-02-22
chore: unified println output for command line utility

septs:


CoelacanthusHex:
According to C standard, libc should treat `\n` as newline and convert it from and to platform newline representation when input and output. It should be a bug of libc, not lpac. Report it to mingw64 instead of leaving a dirty patch in lpac.

PR #199, opened 2025-02-26
feat: error stringify for win32

septs:


driver/apdu/pcsc_win32.c
```diff
@@ -0,0 +1,70 @@
+#ifdef _WIN32
+#include "pcsc_win32.h"
+#include <winscard.h>
+
+static char *pcsc_stringify_error_win32(const DWORD err) {
```
CoelacanthusHex, line 5: Why static? It will make this function only available in this file (strictly, transition unit).

PR #201, opened 2025-02-26
refactor: always use \n for newline and make fuart unbuffered

CoelacanthusHex:
> C89 Standard section 5.2.2 Character display semantics
> \n (new line) Moves the active position to the initial position of the next line.

'\n' in C have including meaning of both '\r' and '\n' on Windows, and
libc SHALL translate them when do input and output. If not, it SHALL be
a bug of libc.

FYI, all mingw64 toolchain code use '\n' directly, so it's impossible to
be caused by '\n', otherwise mingw64 itself will broken. Mingw64 also
set _fmode to _O_TEXT by default.[1] So as MSVC documents said[2], it
should do translate between LF and CRLF.

[1]: https://github.com/mingw-w64/mingw-w64/blob/master/mingw-w64-crt/crt/txtmode.c
[2]: https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/fopen-wfopen

### DIFF
--- a/driver/apdu/at.c
+++ b/driver/apdu/at.c
@@ -26,7 +26,7 @@ static int at_expect(char **response, const char *expected)
         fgets(buffer, AT_BUFFER_SIZE, fuart);
         buffer[strcspn(buffer, "\r\n")] = 0;
         if (getenv("AT_DEBUG"))
-            printf("AT_DEBUG: %s\r\n", buffer);
+            printf("AT_DEBUG: %s\n", buffer);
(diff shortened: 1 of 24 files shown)


septs:
https://stackoverflow.com/a/13286843

AT command, it's use `\r\n`

CoelacanthusHex:
> https://stackoverflow.com/a/13286843
> 
> AT command, it's use `\r\n`

Yeah. I left all \r\n for fuart.

PR #202, opened 2025-03-01
chore: improve pcsc_stringify_error for win32

septs:


CoelacanthusHex (approved):
Should be safe since GCC documented this behavior.

PR #204, opened 2025-03-02
chore: libqmi-glib minimum version constraint

septs:
```console
$ cmake .. -DLPAC_WITH_APDU_QMI=ON
CMake Deprecation Warning at CMakeLists.txt:1 (cmake_minimum_required):
  Compatibility with CMake < 3.10 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
  to tell CMake that the project requires at least <min> but has been updated
  to work with policies introduced by <max> or earlier.


-- Checking for module 'qmi-glib>=1.35.5'
--   Package dependency requirement 'qmi-glib >= 1.35.5' could not be satisfied.
Package 'qmi-glib' has version '1.34.0', required version is '>= 1.35.5'
CMake Error at /usr/share/cmake/Modules/FindPkgConfig.cmake:645 (message):
  The following required packages were not found:

   - qmi-glib>=1.35.5

Call Stack (most recent call first):
  /usr/share/cmake/Modules/FindPkgConfig.cmake:873 (_pkg_check_modules_internal)
  driver/CMakeLists.txt:76 (pkg_check_modules)


-- Configuring incomplete, errors occurred!
```


CoelacanthusHex (approved):
LGTM

PR #207, opened 2025-03-06
CMakeLists.txt: enable CMAKE_EXPORT_COMPILE_COMMANDS by default

yegorich:
This creates a `compile_commands.json` file that can then be used by the `cland` tool.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,6 +11,8 @@ set(CMAKE_C_STANDARD_REQUIRED ON)
 set(LPAC_CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
 list(APPEND CMAKE_MODULE_PATH ${LPAC_CMAKE_MODULE_PATH})
 
+set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
+
 # add_compile_options(-Wall -Wextra -Wpedantic)
 
 # Enable LTO when possible.
@@ -45,13 +47,13 @@ endif()
 
 if(CPACK_GENERATOR)
     set(CPACK_PACKAGE_VENDOR "eSTK.me Group")
-    
+
     set(CPACK_DEBIAN_PACKAGE_MAINTAINER "eSTK.me Group")
     set(CPACK_DEBIAN_PACKAGE_DEPENDS "libc6")
     set(CPACK_DEBIAN_PACKAGE_RECOMMENDS "libcurl, libpcsclite, pcscd")
 
     set(CPACK_DEBIAN_FILE_NAME DEB-DEFAULT)
-    
+
     set(CPACK_RPM_PACKAGE_LICENSE "AGPL-3.0-only AND LGPL-2.0-only")
     set(CPACK_RPM_PACKAGE_AUTOREQ "yes")
     set(CPACK_RPM_PACKAGE_REQUIRES "libcurl, libpcsclite, pcscd")


CMakeLists.txt
```diff
@@ -11,6 +11,10 @@ set(CMAKE_C_STANDARD_REQUIRED ON)
 set(LPAC_CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
 list(APPEND CMAKE_MODULE_PATH ${LPAC_CMAKE_MODULE_PATH})
 
+if(CMAKE_EXPORT_COMPILE_COMMANDS STREQUAL "")
+  set(CMAKE_EXPORT_COMPILE_COMMANDS ON CACHE BOOL "project default" FORCE)
+endif()
```
CoelacanthusHex, line 16: Use `set(CMAKE_EXPORT_COMPILE_COMMANDS ON)` directly. Force overriding the cached variable is not a good idea. The developers should rebuild the build cache by themselves.

yegorich: OK. Fixed.

septs:
@CoelacanthusHex marge?

CMakeLists.txt
```diff
@@ -11,6 +11,10 @@ set(CMAKE_C_STANDARD_REQUIRED ON)
 set(LPAC_CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
 list(APPEND CMAKE_MODULE_PATH ${LPAC_CMAKE_MODULE_PATH})
 
+if(CMAKE_EXPORT_COMPILE_COMMANDS STREQUAL "")
```
CoelacanthusHex, line 14: Please use `NOT DEFINED` instead.

CMakeLists.txt
```diff
@@ -11,6 +11,10 @@ set(CMAKE_C_STANDARD_REQUIRED ON)
...
+  set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
```
septs, line 15: indent problem, other code is 4 spaces, but this is 2 spaces

CMakeLists.txt
yegorich: Do you mean something like that?

```
if(NOT DEFINED CMAKE_EXPORT_COMPILE_COMMANDS)
    set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
endif()
```

This doesn't work.

CMakeLists.txt
yegorich: Fixed.

CMakeLists.txt
yegorich: Or we can omit the guard completely and force compile_commands.json creation. Meson build system makes this by default. It also creates a .gitignore file inside the build directory.

septs: @CoelacanthusHex

CoelacanthusHex: I think it's better to follow Meson to remove guard. `.gitignore` is unnecessary because out-of-source build is recommended.

yegorich: OK. Than I'll send an update

yegorich: Ready

yegorich: @CoelacanthusHex `.gitignore` in the out-of-source folder is really nice. So far, we have `build` entry in the root `.gitignore`. But if you need another build folder `build-cross`, it will immidiatelly "pollute" you git repo. This is why meson automatically creates a `.gitignore` file inside the out-of-source build directory. Because in this case, git just ignores it without an entry inside the root `.gitignore` file. See this [post](https://www.scivision.dev/cmake-auto-gitignore-build-dir/).

CoelacanthusHex (approved):
LGTM

PR #213, opened 2025-03-12
driver/apdu/qmi.c: use calloc instead of malloc

yegorich:
`calloc()` also zeros all elements.

### DIFF
--- a/driver/apdu/qmi.c
+++ b/driver/apdu/qmi.c
@@ -51,7 +51,7 @@ static int libapduinterface_init(struct euicc_apdu_interface *ifstruct)
 {
     struct qmi_data *qmi_priv;
 
-    qmi_priv = malloc(sizeof(struct qmi_data));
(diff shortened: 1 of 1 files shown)


septs:
why cannot use `memset(qmi_priv, 0, sizeof(qmi_priv))`?

yegorich:
We can also use `memset`, but with `calloc` we have just one call instead of two. Let me know which way you prefer.

septs:
see https://github.com/search?q=repo%3Aestkme-group%2Flpac%20memset&type=code

yegorich:
Here is an interesting blog post about `malloc` + `memset` vs `calloc`: https://vorpus.org/blog/why-does-calloc-exist/

yegorich:
Looks like calloc is both safer and more efficient. Hence, we could convert all occurrences of malloc and memset to calloc.

CoelacanthusHex (approved):
LGTM

Issue #215, opened 2025-03-14
qmi_qrtr driver regression


z3ntu:
I was notified that the qmi_qrtr driver doesn't seem to work anymore with the following error:
```
error: open QMI device failed: endpoint hangup
```
After bisecting this issue it leads to https://github.com/estkme-group/lpac/commit/3bde4a1d3dd6ada280ac3b3e53a6b65997ca8f27#diff-662459dc9db81ea57edce076b82be347a8895dd54b6f8a83d52655511dcb3a14R98-R99.

So when changing `QMI_DEVICE_OPEN_FLAGS_PROXY` back to `QMI_DEVICE_OPEN_FLAGS_NONE` it starts working again. This also makes sense because there's no QMI proxy for QRTR, this probably only applies to opening a QMI device through /dev.

https://www.freedesktop.org/software/libqmi/libqmi-glib/latest/QmiDevice.html#QmiDeviceOpenFlags

Obviously this change was done for a purpose to I wanna check with mostly @robimarko how we can resolve this. Using `QMI_DEVICE_OPEN_FLAGS_AUTO` also works for QRTR, but I'm not sure if that would use the proxy if available for the plain QMI backend.

<details>

> driver: apdu: qmi-helpers: support opening QMI device in proxy mode
> 
> Currently, we are using QMI_DEVICE_OPEN_FLAGS_NONE to open the QMI device
> and thus we are requesting exclusive access to the device while we need
> to talk to it.
> 
> This will work fine as long as we are the only thing trying to use that
> QMI device at the same time or the device was not already opened in proxy
> mode.
> 
> This is an issue since ModemManager will open the device in proxy mode so
> that qmicli or other applications can still talk to the same QMI device
> but it will break lpac from trying to use QMI.
> 
> So, lets try and open the device in proxy mode, libqmi will the start the
> qmi-proxy service automatically as its built and installed as part of it.

</details>

If this doesn't work, then we'll probably have to pass the different flags from `qmi_qrtr.c` and `qmi.c` which would also be fine.

robimarko:
@z3ntu Hi there, sorry about breaking QRTR support.

As far as I know `QMI_DEVICE_OPEN_FLAGS_AUTO` won't work since it just means open in QMI or MBIM mode based on the underlying driver, so the only solution would be to pass different flags whether QRTR or plain QMI is being used

z3ntu:
[...]

PR #216, opened 2025-03-15
driver: apdu: qmi*: Pass correct open flag for QRTR driver

z3ntu:
While the QMI driver which goes to /dev/foobar wants to go through the qmi-proxy to not have to exclusively claim the device, the QMI-over-QRTR driver doesn't work with the PROXY flag and just leads to the error

  error: open QMI device failed: endpoint hangup

Fix this by passing the correct flag to qmi_device_open from both APDU drivers.

Closes: #215
Fixes: 3bde4a1 (" driver(APDU): add QMI backend (#131)")

### DIFF
--- a/driver/apdu/qmi.c
+++ b/driver/apdu/qmi.c
@@ -30,7 +30,7 @@ static int apdu_interface_connect(struct euicc_ctx *ctx)
         return -1;
     }
 
-    qmi_device_open_sync(device, qmi_priv->context, &error);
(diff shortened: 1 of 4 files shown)


septs:
why cannot export as env vars?

septs:
is it necessary to export QmiDeviceOpenFlags as env vars?

robimarko:
I dont see why you wouldnt want to use QMI proxy, it wont hurt anything and allows multiple clients to access the QMI

PR #218, opened 2025-03-16
feat: unified env vars style

septs:


driver/apdu/pcsc.c
```diff
@@ -113,9 +114,9 @@ static int pcsc_open_hCard_iter(int index, const char *reader, void *userdata)
     DWORD dwActiveProtocol;
 
     id = 0;
-    if (getenv(INTERFACE_SELECT_ENV))
+    if (getenv(ENV_IFID) != NULL)
     {
-        id = atoi(getenv(INTERFACE_SELECT_ENV));
+        id = (int) strtol(getenv(ENV_IFID), NULL, 10);
```
CoelacanthusHex, line 119: If you will cast to int later, why not use atoi directly?

driver/helpers.c
```diff
@@ -0,0 +1,23 @@
+#include <stdlib.h>
+#include <string.h>
+#include <stdio.h>
+
+#ifdef _WIN32
+int setenv(const char *name, const char *value, const int overwrite) {
+    if (overwrite) goto replace;
```
CoelacanthusHex, line 7: Use structured programming, please.

driver/helpers.c
```diff
@@ -0,0 +1,23 @@
...
+    size_t n = 0;
+    const int errcode = getenv_s(&n, NULL, 0, name);
```
CoelacanthusHex, line 9: It may be better to use `getenv` and `_putenv` to get better consistency since POSIX version of getenv and putenv is not thread-safe.

driver/apdu/pcsc.c
septs: Clang-Tidy: 'atoi' used to convert a string to an integer value, but function will not report conversion errors; consider using 'strtol' instead

septs:
CC @yegorich @z3ntu @lweiss-fairphone @fransklaver

request review the pull request

z3ntu:
Sounds good, thanks for keeping support for the old variables, so that there's no immediate breaking change.

driver/helpers.c
```diff
@@ -0,0 +1,50 @@
...
+}
+#endif
+
+const char *getenv_or_default(const char *name, const char *default_value) {
+    const char *value = getenv(name);
+    if (value == NULL) return default_value;
+    return value;
+}
+
+bool getenv_bool(const char *name, const bool default_value) {
```
CoelacanthusHex, line 25: I suggest it named `getenv_bool_or_default` to inform this function with a default value.

driver/helpers.c
```diff
@@ -0,0 +1,50 @@
...
+    return strcmp(value, "1") == 0 ||
+           strcmp(value, "y") == 0 ||
+           strcmp(value, "Y") == 0 ||
+           strcmp(value, "yes") == 0 ||
+           strcmp(value, "YES") == 0 ||
+           strcmp(value, "true") == 0 ||
+           strcmp(value, "TRUE") == 0;
+}
+
+long getenv_long(const char *name, const long default_value) {
```
CoelacanthusHex, line 37: Maybe we can implement a unified function entry with `_Generic` like std did in `tgmath.h`.

https://en.cppreference.com/w/c/language/generic

driver/helpers.c
```diff
@@ -0,0 +1,50 @@
+#include "helpers.h"
+#include <stdlib.h>
+#include <string.h>
+#include <stdio.h>
+
+#ifdef _WIN32
+int setenv(const char *name, const char *value, const int overwrite) {
+    if (overwrite != 0 && getenv(name) != NULL) return -1;
+    const size_t n = strlen(name) + 1 /* = */ + strlen(value) + 1 /* \0 */;
+    char *env = calloc(n, sizeof(char));
```
CoelacanthusHex, line 10: It might be better to use VLA to reduce the heap allocation cost since setenv is always at the tail of the call chain.

driver/helpers.c
```diff
@@ -0,0 +1,52 @@
...
+const char *getenv_str_or_default(const char *name, const char *default_value) {
+    const char *value = getenv(name);
+    if (value == NULL) return default_value;
+    return value;
+}
+
+bool getenv_bool_or_default(const char *name, const bool default_value) {
+    const char *value = getenv(name);
+    if (value == NULL) return default_value;
+    return strcmp(value, "1") == 0 ||
```
CoelacanthusHex, line 26: In practice, normally all nonzero numbers are true.

fransklaver:
Looks good to me. This generic backend would have prevented at least one bug already. I'm a bit in two minds about the duplication in cases like UIM_SLOT. It feels like it should be a generic apdu setting because regardless of the backend, you should probably be able to select the UIM slot if multiple are available.

I don't think it's a blocking issue though.

CoelacanthusHex (approved):
LGTM

z3ntu:
FYI my lpa-gtk UI is now adjusted to support the new variable for QMI_QRTR: https://codeberg.org/lucaweiss/lpa-gtk/commit/d132d2dc35ddd44204611c89f69be17574cfdb51

PR #223, opened 2025-03-19
Bump minimum CMake version

yegorich:
Fixes https://github.com/estkme-group/lpac/issues/220

PR #224, opened 2025-03-19
cmake: automatically create .gitignore in the build folder

yegorich:
This way, every out-of-source build folder will be ignored.


yegorich:
Depends on https://github.com/estkme-group/lpac/pull/223

yegorich:
Motivation: https://www.scivision.dev/cmake-auto-gitignore-build-dir/

PR #228, opened 2025-03-20
Don't strip the trailing whitespaces for MarkDown files

yegorich:


septs:
I think needs select a markdown formatter, and follow the markdown formatter style.

yegorich:
Without this change, my VIM editor removes the trailing white spaces. And as @CoelacanthusHex pointed out, the white spaces are used to break line in Markdown.

Later, we could add some markdown formatter (uncrustify, cland.format, whatever).

yegorich:
@CoelacanthusHex accodring to [MarkDown syntax](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax), there are three ways to introduce a line break:

- Include two spaces at the end of the first line
- Include a backslash at the end of the first line
- Include an HTML single line break tag at the end of the first line

What about switching to one of the latter choices? Those symbols are visible in all editors. Then, we won't need this PR at all. What do you think?

CoelacanthusHex:
I don't have any special preference for this, so I think the backslash might be a suitable option. It's visible and not noisy. I also suggest adding a space before the backslash to make it easier to distinguish from normal text.

yegorich:
Great. Backslash is my favourite too.

I'm closing this PR and will create another one with adding backslashes.

PR #229, opened 2025-03-20
gbinder_hidl: Fix compiler warnings (and error on older gcc)

mariogrip:


mariogrip:
That was fast :D Thx!

Issue #174, opened 2024-12-03
Segmentation fault for 'lpac driver http' when using QMI backend


AndreySV:
I was just experimenting with lpac and found that is segfaults in following situation:

```console
$ sudo env LPAC_APDU=qmi QMI_DEVICE=/dev/cdc-wdm0 LIBEUICC_DEBUG_APDU=1 ./output/lpac driver
Usage: driver <apdu|http>

$ sudo env LPAC_APDU=qmi QMI_DEVICE=/dev/cdc-wdm0 LIBEUICC_DEBUG_APDU=1 ./output/lpac driver http
Cleaning up leaked APDU channel 1375198244
Segmentation fault
```

CoelacanthusHex:
I have a guess: in qmi driver init code, we allocate memory for qmi_priv
https://github.com/estkme-group/lpac/blob/79034c665ee3bc0e9f9938859916bdfcb2e3a8dd/driver/apdu/qmi.c#L54
But doesn't init it with default value. So there is junk data now. When invoking `lpac driver http`, qmi driver doesn't do any actual thing, so it leaves junk data. When the fini function was invoked, it lookup `lastChannelId` field found non-zero junk data, and then tried to destroy the resource of that channel.
https://github.com/estkme-group/lpac/blob/79034c665ee3bc0e9f9938859916bdfcb2e3a8dd/driver/apdu/qmi_common.c#L155-L163

yegorich:
@CoelacanthusHex the `qmi_apdu_interface_logic_channel_close` is called with the first parameter `ctx` set to `NULL` and in the routine itself, this parameter will be used:

```
void qmi_apdu_interface_logic_channel_close(struct euicc_ctx *ctx, uint8_t channel)
 {
    struct qmi_data *qmi_priv = ctx->apdu.interface->userdata;
```

This produces the segmentation fault.

yegorich:
https://github.com/estkme-group/lpac/pull/213 doesn't fix the issue with closing the QMI channel, but it at least mkaes sure the `lastChannelId` is zero if it wasn't really used.

yegorich:
@septs  the issue with `qmi_apdu_interface_logic_channel_close()` that caused a segfault is still not fixed: 

https://github.com/estkme-group/lpac/blob/79034c665ee3bc0e9f9938859916bdfcb2e3a8dd/driver/apdu/qmi_common.c#L155-L163

[...]

PR #230, opened 2025-03-20
fix: add pragma once

septs:


septs:
@yegorich non-declare `#pragma once`, may be is #174 related-problem?

yegorich:
@septs no. One has nothing to do with the other.  You cannot pass `NULL` to the function that awaits a valid pointer and uses it without checking whether the pointer is `NULL`.

PR #233, opened 2025-03-25
Remove qmi_cleanup routine

yegorich:
`qmi_apdu_interface_logic_channel_close()` called from within `main_fini_euicc()` should already close the channel and invalidate the `qmi_priv->lastChannelId` variable.


yegorich:
Don't pay attention to the changed formatting. I'll fix it. We need `.clang-format` file :-)

yegorich:
Format restored.

PR #238, opened 2025-05-12
driver: apdu: at: don't expect space after "+CGLA:"

vstarodubtsev:
Some devices have a difference in format and do not add a space.

Simcom A7908E:
  AT+CGLA=1,16,"81E2910003BF3C00"
  +CGLA:4,"611A"

  OK


CoelacanthusHex (approved):
Should be safe.

PR #244, opened 2025-06-24
driver: apdu: mbim: sleep while waiting for sim

### LINKED ISSUE #241: Unable to query esim chip info on second slot using mbim
I have a Qualcomm Snapdragon X55 5G modem with two slots, the first one is a physical SIM, the second is eSIM. If I run `lpac chip info` on the first slot, I get an error that the sim isn't ready (I expect this to not work), but then if I run `chip info` on the second slot, which should return data, I get the same error... 

The only way I've been able to get this to work is if I run `chip info` on the second slot, then sleep 1 second, and run it again. I get data back on the second invocation on slot 2.

```
$ export LPAC_APDU=mbim
$ export MBIM_DEVICE=/dev/wwan0mbim0
$ export MBIM_USE_PROXY=1
$ UIM_SLOT=1 lpac chip info; UIM_SLOT=2 lpac chip info; sleep 1; UIM_SLOT=2 lpac chip info

sim did not become available
{"type":"lpa","payload":{"code":-1,"message":"euicc_init","data":""}}
sim did not become available
{"type":"lpa","payload":{"code":-1,"message":"euicc_init","data":""}}
{"type":"lpa","payload":{"code":0,"message":"success","data": ...}}
```

If I remove the sleep, the second invocation fails. I suspect(?) running chip info on a slot that's not eSIM might be tripping up the modem somehow, but really I have no clue 😅 

This is all mimicking the behavior of lpa-gtk, it uses `lpac chip info` to discover slots and drivers to use with lpac.

fransklaver:
When waiting for the sim to become available after potentially switching, we don't actually do much waiting. We're basically only delayed by the speed of the communication here. If the switch doesn't happen quickly enough, this may result in us deciding a sim didn't become available, but subsequent calls will produce usable results.

So introduce a small 50 ms wait time, so we limit the total waiting time to about 1 second. This I think should be enough for the sim to actually become available.

resolve #241

### DIFF
--- a/driver/apdu/mbim.c
+++ b/driver/apdu/mbim.c
@@ -107,6 +107,11 @@ static int select_sim_slot(struct mbim_data *mbim_priv)
         if (is_sim_available(mbim_priv)) {
             return 0;
         }
+        struct timespec ts = {
+            .tv_sec = 0,
+            .tv_nsec = 50000000
+        };
+        nanosleep(&ts, NULL);
     }
 
     fprintf(stderr, "sim did not become available\n");


driver/apdu/mbim.c
```diff
@@ -107,6 +107,10 @@ static int select_sim_slot(struct mbim_data *mbim_priv)
         if (is_sim_available(mbim_priv)) {
             return 0;
         }
+        struct timespec ts;
+        ts.tv_sec = 0;
+        ts.tv_nsec = 100000;
```
CoelacanthusHex, line 112:
```suggestion
        struct timespec ts = {
            .tv_sec = 0,
            .tv_nsec = 100000
        };
```
It can even be rewritten shortly as
```c
nanosleep(&(struct timespec){ .tv_sec = 0, .tv_nsec = 100000 }, NULL);
```
But the readability is not good.
And unfortunately, the C standard replacement `thrd_sleep(3)` is only available since C11.

fransklaver: Thanks, I more or less assumed we couldn't do this.

fransklaver: Nah, I was wrong. The communication introduces significant delay, it just wasn't significant enough. I'll change the length of the wait to the intended value.

fransklaver:
Switched to waits of 50ms, as the ~2 second wait felt too long.

PR #245, opened 2025-07-13
driver: add APDU AT backend support for Win32

Shua-github:
This backend allows selecting the device via the `AT_DEVICE_WIN32` environment variable. It has been tested successfully only on my EC20 device using the `lpac chip info` command, and has not been tested on other devices.


driver/apdu/at_win32.c
```diff
@@ -0,0 +1,348 @@
...
+        at_read_buffer_len -= (line_len + 1);
+
+        line[strcspn(line, "\r")] = 0;
+
+        if (strlen(line) == 0) {
+            continue;
+        }
+
+        if (getenv("AT_DEBUG"))
+            printf("AT_DEBUG_RX: %s\n", line);
```
CoelacanthusHex, line 66: Debug output should be sent to stderr via `fprintf(stderr, "")` so it can be isolated with "true output" easily by users.
Same issue in other places.

driver/apdu/at_win32.c
```diff
@@ -0,0 +1,348 @@
...
+
+static int apdu_interface_connect(struct euicc_ctx *ctx)
+{
+    const char *device;
+    DCB dcb = {0};
+    COMMTIMEOUTS timeouts = {0};
+
+    logic_channel = 0;
+
+    if (!(device = getenv("AT_DEVICE_WIN32")))
```
CoelacanthusHex, line 118: I think we can reuse environment variable name "AT_DEVICE", since apdu/at and apdu/win32 will never be used together.
And add document for it.

driver/apdu/at_win32.c
```diff
@@ -0,0 +1,348 @@
...
+    {
+        device = "COM3";
```
CoelacanthusHex, line 120: Why it's 3? It's not a normal choice.

driver/apdu/at_win32.c
```diff
@@ -0,0 +1,348 @@
...
+    }
+
+    char devname[64];
+    snprintf(devname, sizeof(devname), "\\\\.\\%s", device);
+
+    hComm = CreateFileA(devname,
```
CoelacanthusHex, line 126: Why use A version function on modern Windows?

driver/CMakeLists.txt
```diff
@@ -53,6 +58,12 @@ if(LPAC_WITH_APDU_AT)
     target_sources(euicc-drivers PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/apdu/at.c)
 endif()
 
+if(LPAC_WITH_APDU_AT_WIN32)
+    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -DLPAC_WITH_APDU_AT_WIN32")
+    target_compile_definitions(euicc-drivers PRIVATE LPAC_WITH_APDU_AT_WIN32)
```
CoelacanthusHex, line 63: These two lines are duplicate.

driver/apdu/at_win32.c
Shua-github: On Quectel devices, the default AT port is usually COM3 (at least that's the case with my device), I don't know what the default value should be here.

Shua-github: USB virtual serial ports appear to start at COM3, although I'm not certain.

septs (approved):

driver/apdu/at_win32.c
```diff
@@ -0,0 +1,341 @@
...
+        return -1;
+    }
+    hComm = INVALID_HANDLE_VALUE;
+
+    return 0;
+}
+
+static int libapduinterface_main(int argc, char **argv)
+{
+    return 0;
```
septs, line 324: please use `GetCommPorts` get available COM ports.

https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-getcommports

reference:

https://github.com/estkme-group/lpac/blob/ac221987455a177e91b34c5437c1d0fa67320422/driver/apdu/pcsc.c#L423-L466

septs (changes requested):

septs (approved):

PR #247, opened 2025-08-05
driver: apdu: mbim: make sim behavior more consistent

fransklaver:
When we do not need to switch sims, the output of the program says

    error: no channel response received: Failure

but when we do switch, we say

    sim did not become available

To make this behavior more consistent, don't say anything about the sim status and continue as if we didn't have to switch.


CoelacanthusHex (approved):
LGTM

PR #253, opened 2025-08-10
fix(pcsc): skip reader if meet sharing violation

septs:
It may be this reader is used by another process (e.g. OpenPGP Card), shouldn't abort iteration.

### DIFF
--- a/driver/apdu/pcsc.c
+++ b/driver/apdu/pcsc.c
@@ -125,6 +125,8 @@ static int pcsc_open_hCard_iter(int index, const char *reader, void *userdata)
     const int ret = SCardConnect(pcsc_ctx, reader, SCARD_SHARE_EXCLUSIVE, SCARD_PROTOCOL_T0, &pcsc_hCard, &dwActiveProtocol);
     if (ret != SCARD_S_SUCCESS)
     {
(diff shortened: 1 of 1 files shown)


driver/apdu/pcsc.c
```diff
@@ -125,6 +125,8 @@ static int pcsc_open_hCard_iter(int index, const char *reader, void *userdata)
     const int ret = SCardConnect(pcsc_ctx, reader, SCARD_SHARE_EXCLUSIVE, SCARD_PROTOCOL_T0, &pcsc_hCard, &dwActiveProtocol);
     if (ret != SCARD_S_SUCCESS)
     {
+        // see <https://blog.apdu.fr/posts/2024/12/gnupg-and-pcsc-conflicts-episode-3/>
+        if (ret == SCARD_E_SHARING_VIOLATION) return 0; // skip
```
estkme, line 129: make a env switch for this, you never want run lpac when your card running firmware upgrade

driver/apdu/pcsc.c
```diff
@@ -125,6 +125,8 @@ static int pcsc_open_hCard_iter(int index, const char *reader, void *userdata)
...
         pcsc_error("SCardConnect()", ret);
```
CoelacanthusHex, line 130: Move pcsc_error() before check.

PR #256, opened 2025-08-10
fix(euicc): strip notification address

septs:


estkme:
NG,

trim should done at lpac level, libeuicc should always return "real" value.

estkme:
...even it seems not corret

PS: reopened, fat finger clicked wrong button

PR #259, opened 2025-08-12
fix(mbim): update default UIM slot to 1 for consistency with QMI

septs:


CoelacanthusHex (approved):

fransklaver:
Yup, that seems reasonable enough.

PR #262, opened 2025-08-13
chore: move scripts directory to .github directory

septs:


.editorconfig
```diff
@@ -10,3 +10,4 @@ trim_trailing_whitespace = true
 
 [*.md]
 indent_size = 2
+trim_trailing_whitespace = false
```
CoelacanthusHex, line 13: Please think twice about how to do a hard line break in Markdown before you write this rule...

septs: An existing file already uses this format

```
Line<space><space>
Line
```

CoelacanthusHex: Oh, my eye is broken... I see false as true. :(

PR #263, opened 2025-08-13
feat: strict matching id checking in activation code

septs:
```console
$ ./lpac profile download -a $'LPA:1\$example.com\$test\ntest'
{"type":"progress","payload":{"code":0,"message":"es10b_cancel_session","data":"example.com"}}
{"type":"progress","payload":{"code":0,"message":"es9p_cancel_session","data":"example.com"}}
{"type":"lpa","payload":{"code":-1,"message":"matching_id","data":"invalid format, contains character not alphanumeric or dash"}}
```

### DIFF
--- a/src/applet/profile/download.c
+++ b/src/applet/profile/download.c
@@ -1,6 +1,7 @@
 #include "download.h"
 #include <stdio.h>
 #include <stdlib.h>
+#include <stdbool.h>
 #include <unistd.h>
 #include <string.h>
 #include <getopt.h>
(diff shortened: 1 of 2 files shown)


src/applet/profile/download.c
```diff
@@ -37,6 +40,15 @@ char *strsep(char **stringp, const char *__delim) {
 }
 #endif
 
+static bool is_strict_matching_id(const char *token) {
+    const size_t n = strlen(token);
+    for (int i = 0; i < n; i++) {
+        if (isalnum(token[i]) || token[i] == '-') continue;
```
CoelacanthusHex, line 46: Don't use `isalnum()`, it will contain other characters when the locale is not "C.UTF-8".

CoelacanthusHex: Or add `setlocale(LC_ALL, "C.UTF-8");` to https://github.com/estkme-group/lpac/blob/f043c01a0028ee4eb0ba374dd771cb90b4770427/src/main.c#L166, and then you can use not `*_l` version everywhere except libeuicc.

CoelacanthusHex (approved):

PR #270, opened 2025-08-15
refactor(lpac): use __attribute__((cleanup)) to do SBRM

CoelacanthusHex:
Ensure all resources will be destroyed safely and correctly without manually releasing (whatever break/return/goto/exception happened). systemd has used it for a long time, and the Linux kernel also chose it 1 year ago.


utils/lpac/utils.h
```diff
@@ -0,0 +1,59 @@
+#pragma once
+#include <cjson/cJSON.h>
+#include <euicc/es9p.h>
+#include <euicc/es10b.h>
+#include <euicc/es10c.h>
+#include <stdbool.h>
+#include <stdlib.h>
+
+#define HTTP_ENV_NAME(DRIVER, NAME) "LPAC_HTTP_" #DRIVER "_" #NAME
```
septs, line 9: https://github.com/estkme-group/lpac/blob/abad55dd4d13b406e74834f215a4fc00e07d1124/src/main.c#L33-L34

move this a define to utils.h? share the string.

CoelacanthusHex: No. It's only used by `src/main.c`, not generic.

driver/CMakeLists.txt
```diff
@@ -18,10 +18,9 @@ if(LPAC_DYNAMIC_DRIVERS)
 else()
     add_library(euicc-drivers STATIC ${DIR_INTERFACE_SRCS})
 endif()
-target_link_libraries(euicc-drivers euicc cjson-static)
+target_link_libraries(euicc-drivers euicc cjson-static lpac_utils)
```
septs, line 21:
```suggestion
target_link_libraries(euicc-drivers euicc cjson-static lpac-utils)
```

i think this style needs to be consistent

utils/lpac/utils.h
septs: https://github.com/estkme-group/lpac/blob/abad55dd4d13b406e74834f215a4fc00e07d1124/driver/driver.c#L118-L119

septs (approved):

euicc/euicc.h
```diff
@@ -4,6 +4,10 @@
 #include "interface.h"
 #include "es10b.h"
 
+#ifdef interface
+#  undef interface
```
septs, line 8: add some comment for this `#undef`?

PR #279, opened 2025-08-16
fix: getenv with bool

septs:
CleanShot 2025-08-16 at 12 28 13@2x

https://github.com/estkme-group/lpac/blob/3f0156a549f5de9e0990f709005364d03c448f72/utils/lpac/utils.h#L45-L50

needs type cast as bool

### DIFF
--- a/driver/apdu/at.c
+++ b/driver/apdu/at.c
@@ -48,7 +48,7 @@ static int at_expect(char **response, const char *expected) {
     while (1) {
         fgets(buffer, AT_BUFFER_SIZE, fuart);
         buffer[strcspn(buffer, "\r\n")] = 0;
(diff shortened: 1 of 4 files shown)


CoelacanthusHex:
That's why GCC doesn't add an explicit type cast in stdbool.h and doesn't warn about such an implicit cast event with `-Wconversion`.
> An object declared as type _Bool is large enough to store the values 0 and 1

C99 6.2.5.2

septs:
```c
#include <stdio.h>
#include <stdbool.h>

#define EVALUATE(X) _Generic((X), _Bool : "boolean", default : "not boolean")

int main(void) {
    printf("%s\n", EVALUATE(true));

    return 0;
}
```

Results:

```console
$ gcc test.c
$ ./a.out
not boolean
```

PR #284, opened 2025-08-17
refactor(driver/at): fully rewrite

### LINKED ISSUE #138: AT command APDU driver violates 3GPP TS 27.007 and will not work on e.g. MC7455

septs:
resolves #138


septs:
@Shua-github request review

septs:
needs more tests

PR #304, opened 2025-08-22
feat(src): handle SIGINT signal to trigger cleanup

septs:


src/main.c
```diff
@@ -148,10 +149,32 @@ static char **warg_to_arg(const int wargc, wchar_t **wargv) {
...
 int main(int argc, char **argv) {
     int ret = 0;
 
     setlocale(LC_ALL, "C.UTF-8");
+    atexit(main_fini_euicc);
+
+    // SIGINT handler
+    {
+#ifdef WIN32
+        signal(SIGINT, signal_handler);
```
CoelacanthusHex, line 168: In `signal(2)`, it said 
> Portability
> The only portable use of signal() is to set a signal's disposition to SIG_DFL or SIG_IGN.  The semantics when using signal() to establish a signal handler vary across systems (and POSIX.1 explicitly permits this variation); do not use it for this purpose.

And it's not thread-safe.
But there is no alternative on Windows, so it's impossible to fix it...

PR #309, opened 2025-08-24
refactor(cmake): allow version from environ

septs:
```bash
cmake -DLPAC_VERSION=v2.3.1
```

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -4,7 +4,7 @@ else()
     set(RPATH_BINARY_PATH "$ORIGIN")
 endif()
 
-set(LPAC_VERSION "" CACHE STRING "Set version string for lpac")
+set(LPAC_VERSION "" STRING "Set version string for lpac")
 
 aux_source_directory(${CMAKE_CURRENT_SOURCE_DIR} DIR_LPAC_SRCS)
 


cmake/git-version.cmake
```diff
@@ -1,25 +1,27 @@
...
-    WORKING_DIRECTORY ${SRC_DIR}
-    OUTPUT_VARIABLE GIT_DESCRIBE_VERSION
-    RESULT_VARIABLE GIT_DESCRIBE_ERROR_CODE
-    OUTPUT_STRIP_TRAILING_WHITESPACE
-  )
-  if(NOT GIT_DESCRIBE_ERROR_CODE)
-    set(LPAC_VERSION ${GIT_DESCRIBE_VERSION})
-  endif()
-endif()
+if (DEFINED ENV{LPAC_VERSION})
```
CoelacanthusHex, line 3: It's better to just use a CMake variable and when build use -DLPAC_VESION=xxx.

CoelacanthusHex:
I think we should remove all git version entirely, or just append revisions (like `.r121`) to the version only when git and .git are available and not fit into a tag.

septs:
> I think we should remove all git version entirely, or just append revisions (like `.r121`) to the version only when git and .git are available and not fit into a tag.

Please create a new pr resolve this problem

PR #312, opened 2025-08-25
fix(main): ensure lpac driver list skips init

Shua-github:
Before the repair:  
```
[root@XiaPuPC output]# ./lpac driver list
SCardEstablishContext() failed: 8010001D (Unknown error: 0xFFFFFFFF8010001D)
APDU driver init failed
```  
After the repair:  
```
[root@XiaPuPC output]# ./lpac driver list
{"type":"driver","payload":{"LPAC_APDU":["pcsc","at","stdio"],"LPAC_HTTP":["curl","stdio"]}}
```

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -34,16 +34,8 @@
 #define ES10X_MSS_MAX_VALUE 255
 
 static const struct applet_entry *driver_applets[] = {
-    &(struct applet_entry){
-        .name = "apdu",
-        .main = euicc_driver_main_apdu,
-        .skip_init_euicc = true
-    },
-    &(struct applet_entry){
-        .name = "http",
-        .main = euicc_driver_main_http,
-        .skip_init_euicc = true
-    },
+    &(struct applet_entry){.name = "apdu", .main = euicc_driver_main_apdu, .skip_init_euicc = true},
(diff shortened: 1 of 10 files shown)


septs (approved):
Wait for @CoelacanthusHex

src/main.c
septs, line 1: extract driver subcommand to new file, e.g: `applet/driver.c`

septs (approved):
move

CoelacanthusHex (dismissed):
It works. But I think it's more elegant to explicitly do euicc_driver_init() and main_euicc_init() when needed by the applet itself. It's more flexible and avoids introducing two branch in the hot path.

CoelacanthusHex:
> > It works. But I think it's more elegant to explicitly do euicc_driver_init() and main_euicc_init() when needed by the applet itself. It's more flexible and avoids introducing two branch in the hot path.
> 
> Should we modify the final applet?

Sorry, maybe my NLP function is poor... But what do you mean?

Shua-github:
> It works. But I think it's more elegant to explicitly do euicc_driver_init() and main_euicc_init() when needed by the applet itself. It's more flexible and avoids introducing two branch in the hot path.

I tried implementing it, but the result is not very elegant; I’m looking for another approach.

PR #318, opened 2025-09-01
feat(euicc): expose logger fp

septs:


euicc/logger.c
```diff
@@ -0,0 +1,74 @@
...
+
+#define BASH_RED 31
+#define BASH_GREEN 32
+#define BASH_CLEAR 0
+
+static void bash_colorize(FILE *fp, const int color) {
+#if _WIN32
+    // disable bash colorize on non-unix platform
+    (void)fp;
+    (void)color;
```
CoelacanthusHex, line 14: Use `_isatty()` to detect ConPTY on Windows.

euicc/logger.c
```diff
@@ -0,0 +1,74 @@
...
+}
+
+inline void euicc_apdu_request_print(FILE *fp, const struct apdu_request *req, const uint32_t req_len) {
+    if (fp == NULL)
+        return;
+    bash_colorize(fp, BASH_GREEN);
+    fprintf(fp, "[DEBUG] [APDU] [TX] CLA: %02X, INS: %02X, P1: %02X, P2: %02X, Lc: %02X, Data:", req->cla, req->ins,
+            req->p1, req->p2, req->length);
+    euicc_hex_print(fp, req->data, req_len - sizeof(struct apdu_request));
+    bash_colorize(fp, BASH_CLEAR);
```
CoelacanthusHex, line 33: Prefer form like
```
fprintf(fp, COLOR_GREEN "[DEBUG] [APDU] [TX] CLA: %02X, INS: %02X, P1: %02X, P2: %02X, Lc: %02X, Data:" COLOR_RESET, req->cla, req->ins,
            req->p1, req->p2, req->length);
```

euicc/logger.c
```diff
@@ -0,0 +1,74 @@
+#include "logger.h"
+
+#include <string.h>
+#include <unistd.h>
+
+#define BASH_RED 31
+#define BASH_GREEN 32
+#define BASH_CLEAR 0
+
+static void bash_colorize(FILE *fp, const int color) {
```
CoelacanthusHex, line 10: Don't use word "bash", it's not a bash-specific feature.

euicc/logger.c
septs: Windows needs use `SetConsoleMode` api set `ENABLE_VIRTUAL_TERMINAL_INPUT`, but in CI not found `consoleapi.h` file 

https://learn.microsoft.com/en-us/windows/console/setconsolemode

septs: https://stackoverflow.com/a/52608157

euicc/logger.c
CoelacanthusHex: No. Just call it `colorize`.
And, VT100 is a monochrome terminal.

euicc/logger.c
```diff
@@ -0,0 +1,74 @@
...
+
+inline void euicc_apdu_request_print(FILE *fp, const struct apdu_request *req, const uint32_t req_len) {
+    if (fp == NULL)
+        return;
+    colorize(fp, COLOR_GREEN);
+    fprintf(fp, "[DEBUG] [APDU] [TX] CLA: %02X, INS: %02X, P1: %02X, P2: %02X, Lc: %02X, Data:", req->cla, req->ins,
+            req->p1, req->p2, req->length);
+    euicc_hex_print(fp, req->data, req_len - sizeof(struct apdu_request));
+    colorize(fp, COLOR_CLEAR);
+    fputc('\n', fp);
```
CoelacanthusHex, line 34: If you can do output in one time, don't split it into multiple statements. Otherwise, it may be interleaved in a multithreaded environment or in a shared fd environment. Since libeuicc is a library, we SHALL consider this issue.

CoelacanthusHex: You can construct the whole string using snprintf, and output it using one call.

septs: CleanShot 2025-09-07 at 23 02 22@2x

PR #319, opened 2025-09-02
fix(src): case-insensitive activation code scheme

septs:
```
./lpac profile download -a 'lpa:1$...$...'
```


CoelacanthusHex:
Is there a specification defining this string?

septs:
> Is there a specification defining this string?

> c

from https://www.gsma.com/solutions-and-impact/technologies/esim/wp-content/uploads/2020/06/SGP.22-v2.2.2.pdf#page=112

PR #320, opened 2025-09-03
refactor(src): applets

septs:
supersede https://github.com/estkme-group/lpac/pull/312


septs:
@Shua-github request review and test it

PR #327, opened 2025-09-06
feat(cmake): static link cjson by default

septs:
```
ld: warning: ignoring file '/opt/homebrew/Cellar/cjson/1.7.18/lib/libcjson.1.7.18.dylib': found architecture 'arm64', required architecture 'x86_64'
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,7 +8,7 @@ project (lpac
 set(CMAKE_C_STANDARD 99)
 set(CMAKE_C_STANDARD_REQUIRED ON)
 
-option(USE_SYSTEM_DEPS "Use system-wide installed dependencies (cJSON)" OFF)
+option(USE_SYSTEM_DEPS "Use system-wide installed dependencies" OFF)
 
 set(LPAC_CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
 list(APPEND CMAKE_MODULE_PATH ${LPAC_CMAKE_MODULE_PATH})


septs:
CC @qaqland

CoelacanthusHex (changes requested):
I think that's an issue of homebrew, it shouldn't add the pkgconfig path of the wrong architecture into the default search path.

septs:
> I think that's an issue of homebrew, it shouldn't add the pkgconfig path of the wrong architecture into the default search path.

https://github.com/estkme-group/lpac/blob/5c5f5c1ef6927f5fa583efadc24523f920b92521/CMakeLists.txt#L34-L36

from https://cmake.org/cmake/help/latest/variable/CMAKE_OSX_ARCHITECTURES.html

`CMAKE_OSX_ARCHITECTURES` requires build arm64 and x86_64 architectures

CoelacanthusHex:
> > I think that's an issue of homebrew, it shouldn't add the pkgconfig path of the wrong architecture into the default search path.
> 
> https://github.com/estkme-group/lpac/blob/5c5f5c1ef6927f5fa583efadc24523f920b92521/CMakeLists.txt#L34-L36
> 
> from [cmake.org/cmake/help/latest/variable/CMAKE_OSX_ARCHITECTURES.html](https://cmake.org/cmake/help/latest/variable/CMAKE_OSX_ARCHITECTURES.html)
> 
> `CMAKE_OSX_ARCHITECTURES` requires build arm64 and x86_64 architectures

So or it should be a bug of CMake. But it shouldn't be a bug of lpac's CMake: it just use the standard way to find dependency.

CoelacanthusHex:
> CleanShot 2025-09-07 at 01 16 26@2x

It's harmless, just ignore it. Or add `QUIET` in find_package to suppress it.

PR #328, opened 2025-09-06
fix(src/jprint): resolve heap corruption

Shua-github:


CoelacanthusHex (changes requested):
Please use SBRM.

CoelacanthusHex:
Are you sure this can fix the crash?
https://github.com/DaveGamble/cJSON/blob/master/cJSON.c#L163
It seems cJSON uses malloc() on Windows as well, so cJSON_free() is the same thing as free(), except it will set the pointer to NULL.

Shua-github:
> Are you sure this can fix the crash? https://github.com/DaveGamble/cJSON/blob/master/cJSON.c#L163 It seems cJSON uses malloc() on Windows as well, so cJSON_free() is the same thing as free(), except it will set the pointer to NULL.

Yes, on my PC.

Shua-github:
[lpac.exe.31268.dmp](https://github.com/user-attachments/files/22190113/lpac.exe.31268.dmp)

CoelacanthusHex:
> > I have this question because we use free() to release `jstr` for a long time. I want to know why it never crashed before.
> 
> I guess it's caused by the dynamic library; it didn’t crash before when I was using static linking.

I asked a Windows expert yesterday, and he told me the reason: the static library cjson built by the original CMakeLists.txt may statically link a CRT, so the malloc/free seen by cjson and lpac is not the same thing. Before https://github.com/estkme-group/lpac/commit/1cc909076945b3b7ec1090e345eae979b3e76f7d, we use our own CMakeLists.txt to build static library cjson, so there is no issue.

PR #339, opened 2025-09-13
feat(cmake): restrict C standard to C99 for euicc target

septs:


euicc/CMakeLists.txt
```diff
@@ -8,6 +8,7 @@ endif()
 target_link_libraries(euicc ${CJSON_LIBRARY} cjson-ext)
 target_include_directories(euicc PUBLIC $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/..>)
 target_compile_options(euicc PRIVATE -Wall -Wextra -Wvla)
+set_target_properties(euicc PROPERTIES C_STANDARD 99 C_STANDARD_REQUIRED ON C_EXTENSIONS ON)
```
CoelacanthusHex, line 11: Please wrap lines for readability.

CoelacanthusHex:
It has been covered by global CMAKE_C_STANDARD and CMAKE_C_STANDARD_REQUIRED in `/CMakeLists.txt`.

CoelacanthusHex:
> > It has been covered by global CMAKE_C_STANDARD and CMAKE_C_STANDARD_REQUIRED in `/CMakeLists.txt`.
> 
> No, you can try to modify `C_STANDARD 99` to `C_STANDARD 90` and then compile

It's expected. Because property has higher precedence than CMAKE_C_STANDARD. If you don't declare C_STANARD in a target, CMake will initialize it with CMAKE_C_STANDARD. If you declare C_STANARD in a target explicitly, it will do nothing. So you do a noop: C_STANDARD of all targets has been set to 99, just setting it again is meaningless.

PR #340, opened 2025-09-13
fix(stdio): use before check

septs:


driver/apdu/stdio.c
```diff
@@ -144,11 +143,11 @@ static int json_response(int *ecode, uint8_t **data, uint32_t *data_len) {
...
 err:
     fret = -1;
-    free(*data);
-    if (data) {
+    if (data != NULL) {
+        free(*data);
         *data = NULL;
     }
-    if (data_len) {
+    if (data_len != NULL) {
```
CoelacanthusHex, line 150: Why will a length be NULL?

septs: https://github.com/estkme-group/lpac/blob/89f901d4302816d40540c06b1ac5d60576700faf/driver/apdu/stdio.c#L82

define is `uint32_t *data_len`

CoelacanthusHex (approved):

PR #341, opened 2025-09-13
refactor(cmake): remove cjson-ext from libeuicc

septs:


CoelacanthusHex (changes requested):
There SHALL be a comparison to inform it has enough improvment in blob size to cancel out the impact on readability and simplicity

septs:
@CoelacanthusHex i think reduce dependencies for better integration

CoelacanthusHex:
> @CoelacanthusHex i think reduce dependencies for better integration

No. It won't. If you use CMake, there is no extra operation needed to do for cjson-ext. If you don't use another build system, you will need to do all the things; cjson-ext will be a very tiny part of it. Don't guess something will give an improvement, you need evidence.

PR #344, opened 2025-09-13
feat: bring dynamic driver back

CoelacanthusHex:
Currently, only supports Linux; the driver SHALL be named as
driver_TYPE_NAME without "lib" prefix, like driver_apdu_pscs.so.
lpac will load drivers from `RUNPATH // drivers`.


CoelacanthusHex:
I can't reproduce https://github.com/estkme-group/lpac/actions/runs/17703912331/job/50313041173?pr=344 on the local machine. I added a workaround for it, but the real reason still needs more investigation. Maybe related to compiler version.

septs:
I think needs provide built in driver option, e.g:

- PC/SC driver on Darwin
- WinHTTP driver on Windows
- Stdio Driver on All Platforms

CoelacanthusHex:
> I think needs provide built in driver option, e.g:
> 
>     * cURL driver on Darwin
> 
>     * PC/SC driver on Darwin
> 
>     * WinHTTP driver on Windows
> 
>     * Stdio Driver on All Platforms

It will make things more complex.
1. Which one will be built-in? Someone has one opinion, others have another. After implementing the dynamic driver, we no longer need to worry whether it will make LPAC completely unusable; ship all supported drivers, let the user choose which one he wants to use.
2. If there is a built-in driver, and there is a dynamic driver with the same name, which one should be used? If we disallow the mixed usage, the logic of lpac driver pick will be more understandable.
3. In my opinion, the built-in driver and the dynamic driver shipped by default are no different for users. They all can get OOBE through both ways. No normal user cares about the single executable file release, nobody!

septs:
If can per-driver owned standalone CMakeLists.txt file, I think better.

driver/driver.c
```diff
@@ -71,7 +87,106 @@ static const struct euicc_driver *_driver_http = NULL;
...
+    strcat(LPAC_DRIVER_HOME, "/drivers");
+
+    DIR *driver_dir = opendir(LPAC_DRIVER_HOME);
+    if (driver_dir == NULL) {
+        return true;
+    }
+    struct dirent *entry;
+    while ((entry = readdir(driver_dir)) != NULL) {
+        if ((entry->d_type & (DT_LNK | DT_REG | DT_UNKNOWN))
+            && (ends_with(entry->d_name, ".so") || ends_with(entry->d_name, ".dll"))) {
```
septs, line 149: Darwin ends with is `.dylib`

driver/driver.c
```diff
@@ -71,7 +87,106 @@ static const struct euicc_driver *_driver_http = NULL;
...
+            struct euicc_driver *driver = dlsym(handle, ifn);
+            if (driver == NULL) {
+                return false;
+            }
+            struct euicc_drivers_list *tmp = (struct euicc_drivers_list *)calloc(1, sizeof(struct euicc_drivers_list));
+            tmp->driver = driver;
+            list_add_tail(&tmp->list, &dynamic_drivers);
+        }
+    }
+    closedir(driver_dir);
```
septs, line 174: SBRM?

CoelacanthusHex:
> If can per-driver owned standalone CMakeLists.txt file, I think better.

Unless we drop the old built-in driver mechanism, it will make CMake more complex and harder to maintain

driver/driver.c
```diff
@@ -1,65 +1,88 @@
+#define _GNU_SOURCE
```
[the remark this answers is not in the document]
septs: please add some comment for this line code

driver/driver.c
```diff
@@ -71,7 +95,184 @@ static const struct euicc_driver *_driver_http = NULL;
...
+        return true;
+    }
+#    if defined(_WIN32)
+    const char *suffix = ".dll";
+#    elif defined(__APPLE__) && defined(__MACH__)
+    const char *suffix = ".dylib";
+#    elif defined(__unix__) || defined(__unix)
+    const char *suffix = ".so";
+#    else
+#        error "Unsupported platform."
```
septs, line 231: may be needs update documenation, add supported platform list

CoelacanthusHex: The platform list is hard to list... And I can't imagine which platform will reach this line, I leave it just for completeness...

driver/driver.c
```diff
@@ -71,7 +98,189 @@ static const struct euicc_driver *_driver_http = NULL;
...
+    return strdup(strtok(runpath1, ":"));
+}
+
+#if defined(__unix__) || defined(__unix)
+static bool init_dynamic_driver_list() {
+    _cleanup_free_ char *LPAC_DRIVER_HOME = get_first_runpath(get_runpath());
+    if (LPAC_DRIVER_HOME == NULL)
+        return false;
+    if (!strcmp(LPAC_DRIVER_HOME, "$ORIGIN")) {
+        free(LPAC_DRIVER_HOME);
```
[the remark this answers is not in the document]
septs: add comment for this line?

driver/CMakeLists.txt
```diff
@@ -14,128 +15,162 @@ cmake_dependent_option(LPAC_WITH_HTTP_WINHTTP "Build HTTP WinHTTP interface (Win
...
 target_include_directories(euicc-drivers PUBLIC ${CMAKE_CURRENT_SOURCE_DIR})
 
 target_sources(euicc-drivers PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/apdu/stdio.c)
 target_sources(euicc-drivers PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/http/stdio.c)
 
+if ((LPAC_DYNAMIC_DRIVERS AND LPAC_DYNAMIC_DRIVERS_SOLID) OR NOT LPAC_DYNAMIC_DRIVERS)
+    set(DRIVER_BUILD_TYPE OBJECT)
+else()
+    set(DRIVER_BUILD_TYPE SHARED)
+endif()
```
septs, line 32:
```suggestion
set(DRIVER_BUILD_TYPE SHARED)
if ((LPAC_DYNAMIC_DRIVERS AND LPAC_DYNAMIC_DRIVERS_SOLID) OR NOT LPAC_DYNAMIC_DRIVERS)
    set(DRIVER_BUILD_TYPE OBJECT)
endif ()
```

septs:
https://github.com/estkme-group/lpac/blob/c0ab7f59bd08008c63a48321d1a12a7810f4bb60/CMakeLists.txt#L54-L68

cpack needs rewrite?

PR #345, opened 2025-09-15
refactor(src): strict type conversion constraint

septs:


src/main.c
```diff
@@ -152,13 +152,13 @@ void main_fini_euicc(void) {
...
 static char **warg_to_arg(const int wargc, wchar_t **wargv) {
-    char **argv = malloc(wargc * sizeof(char *));
+    char **argv = calloc((size_t)wargc, sizeof(char *));
     if (argv == NULL) {
         return NULL;
     }
     for (int i = 0; i < wargc; ++i) {
         const int size = WideCharToMultiByte(CP_UTF8, 0, wargv[i], -1, NULL, 0, NULL, NULL);
-        argv[i] = malloc(size);
+        argv[i] = malloc((size_t)size);
```
CoelacanthusHex, line 161: Move type cast to L160?

septs: https://learn.microsoft.com/en-us/windows/win32/api/stringapiset/nf-stringapiset-widechartomultibyte

`WideCharToMultiByte` returns type is `int`

CoelacanthusHex: Yeah. I mean just cast at the return of WideCharToMultiByte(). `malloc((size_t)size)`'s readabilty is a bit poor.

septs: now is minimize changes

PR #349, opened 2025-09-17
refactor(cmake): use FILE_SET to do headers handling better

CoelacanthusHex:

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -4,15 +4,35 @@ else()
     set(RPATH_BINARY_PATH "$ORIGIN")
 endif()
 
-aux_source_directory(${CMAKE_CURRENT_SOURCE_DIR} DIR_LPAC_SRCS)
-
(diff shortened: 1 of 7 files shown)


CoelacanthusHex:
> select a CMakeLists.txt file formatter

Should be done in separate PR. It will introduce a commit with only format operation, and need keep standalone.

src/CMakeLists.txt
```diff
@@ -4,16 +4,45 @@ else()
...
-
-add_executable(lpac ${DIR_LPAC_SRCS} lpac.manifest)
-target_link_libraries(lpac euicc-drivers lpac-utils)
-target_include_directories(lpac PUBLIC $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>)
+add_executable(lpac lpac.manifest)
+target_link_libraries(lpac PRIVATE euicc euicc-drivers lpac-utils)
+target_sources(
+    lpac
+    PRIVATE
+        applet/chip/defaultsmdp.c
```
septs, line 12: why needs expand file list?

septs (approved):

PR #353, opened 2025-09-20
feat(utils): add path_concat()

CoelacanthusHex:


utils/lpac/utils.c
```diff
@@ -145,3 +145,13 @@ char **merge_array_of_str(char *left[], char *right[]) {
     memcpy(merged + left_len, right, right_len * sizeof(char *));
     return merged;
 }
+
+char *path_concat(const char *restrict a, const char *restrict b) {
```
septs, line 149: i think needs supports variable length arguments

CoelacanthusHex: The situation of concatenating multiple paths (n>=3) is rare, and variable argument support will make this function far more complex (and harder for the compiler to optimize).

septs (approved):

PR #377, opened 2025-09-25
refactor(driver): make driver mechanism full dynamic

CoelacanthusHex:


driver/euicc-driver-loader.c
septs, line 1: split the file

- driver-loader.h
- driver-loader-win32.c
- driver-loader-darwin.c
- driver-loader-linux.c
- driver-loader-unix.c
- etc

septs (changes requested):

driver/euicc-driver-loader.c
septs: euicc-driver-loader.c calling driver-loader.c generalized api

PR #382, opened 2025-10-03
Add WWAN modem & port checking and setting (including for Quectel Modems)

Josef-Ouano:
Add WWAN module, port checking and setting for the following Quectel Modems: 
RM520NGL
EM160RGL
EM05G
EM061kGL

Summary:
- This merge request aims to do the following:
1. Automate the checking of WWAN module AT port if no specific port is set (this applies to above Quectel modems).
 If no appropriate port is detected, default AT port is set.
2. Added checking for WWAN module. 
Purpose is to disable "setbuf(fuart, NULL)" if Quectel EM061k-GL wwan modem is detected, as this causes hangup for this specific modem.

### DIFF
--- a/driver/apdu/at_unix.c
+++ b/driver/apdu/at_unix.c
@@ -16,6 +16,17 @@ static FILE *fuart;
 static int logic_channel = 0;
 static char *buffer;
 
+enum WWAN_MODULE
+{
+    RM520N,
+    EM160R,
+    EM05G,
+    EM061K,
+    NOT_SUPPORTED
+};
+
+#define AT_PORT_DEFAULT "/dev/ttyUSB0"
+
 static void enumerate_serial_device_linux(cJSON *data) {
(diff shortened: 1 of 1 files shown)


septs:
REJECTED, I don't think there should be unique support for specific hardware

CoelacanthusHex (changes requested):
Please format the code with clang-format first.

CoelacanthusHex (changes requested):

Josef-Ouano:
> > Me and my colleague did have a discussion regarding this, if this request is not applicable, for it to be a separate app. I think it may be the possible route for this.
> 
> If accepting this pull request means we have to provide compatibility guarantees for lpac with some specific modems, i think we cannot provide such guarantees.
> 
> This indirectly requires the AT driver to provide compatibility guarantees for some specific modems.

I understand. It looks like doing this as a separate app is the feasible way for these kind of modifications.

PR #397, opened 2025-10-20
docs(docs/at): add caveats

Shua-github:
For equipment from some manufacturers, SIM card sleep must be disabled by issuing the `AT+ECSIMSLEEP=0` command before executing SIM-related commands such as `AT+CSIM`.

### DIFF
--- a/docs/backends/at.md
+++ b/docs/backends/at.md
@@ -29,9 +29,10 @@ If you need to allow **non-root** users to access serial ports, read the followi
 3. <https://wiki.archlinux.org/title/Working_with_the_serial_console#Connect_using_a_terminal_emulator_program>
    (Arch Linux)
 
-> [!NOTE]
->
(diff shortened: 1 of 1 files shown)


docs/backends/at.md
```diff
@@ -29,6 +29,10 @@ If you need to allow **non-root** users to access serial ports, read the followi
 3. <https://wiki.archlinux.org/title/Working_with_the_serial_console#Connect_using_a_terminal_emulator_program>
    (Arch Linux)
 
+> [!NOTE]
+>
+> If you encounter the `+CME ERROR: 3` issue, please try sending `AT+ECSIMSLEEP=0` to the device's AT port, then try again.
```
septs: Need to declare this is a non-standard command of OEM

This is not a general command, and OEM may use a different command name.

septs (approved):

CoelacanthusHex (approved):

PR #399, opened 2025-10-27
fix(driver): 'json_request' interface was changed in b23f0e4

vstarodubtsev:
After a refactoring #270, the skdio driver stopped working correctly.
It turned out that the json_request function previously returned an int and zero in the positive case.
After the refactoring, the positive result was true—the behavior was reversed.
The patch only changes the function's usage, although the style should have returned an int, as in all other driver functions.

### DIFF
--- a/driver/apdu/stdio.c
+++ b/driver/apdu/stdio.c
@@ -158,7 +158,7 @@ exit:
 static int apdu_interface_connect(struct euicc_ctx *ctx) {
     int ecode;
 
-    if (json_request("connect", NULL, 0)) {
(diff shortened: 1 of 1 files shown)


CoelacanthusHex:
> The patch only changes the function's usage, although the style should have returned an int, as in all other driver functions.

Yeah. The mixed style is mainly because we originally wrote this project with C89, where no bool type. When we switch to C99 (with some C11 features), we don't perform a complete refactor.

septs (approved):
LGTM

PR #418, opened 2026-02-26
Allow selecting SIM slot using environment variable in gbinder driver

juanro49:
This solve https://github.com/estkme-group/lpac/issues/413

### DIFF
--- a/driver/apdu/gbinder_hidl.c
+++ b/driver/apdu/gbinder_hidl.c
@@ -9,6 +9,7 @@
 #include <stdlib.h>
 
 #define ENV_DEBUG APDU_ENV_NAME(GBINDER, DEBUG)
+#define ENV_UIM_SLOT APDU_ENV_NAME(GBINDER, UIM_SLOT)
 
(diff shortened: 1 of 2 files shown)


rinigus:
Ping. Please review this PR - we would like to have this option in Sailfish

driver/apdu/gbinder_hidl.c
```diff
@@ -273,6 +283,7 @@ static int apdu_interface_transmit(struct euicc_ctx *ctx, uint8_t **rx, uint32_t
 
 static int libapduinterface_init(struct euicc_apdu_interface *ifstruct) {
     set_deprecated_env_name(ENV_DEBUG, "GBINDER_APDU_DEBUG");
+    set_deprecated_env_name(ENV_UIM_SLOT, "UIM_SLOT");
```
CoelacanthusHex, line 286: For environment variable newly added, don't use set_deprecated_env_name() which is used for backward compatible with old env variables.

PR #420, opened 2026-03-04
feat(euicc): return the ICCID in the result of load_bound_profile_package

maaarghk:
i found it helpful for the ICCID to be readily accessible at the end of a successful transaction, for tying it together with the incoming activation code and doing an automatic "change nickname" operation.

It feels like `struct es10b_notification_metadata_list` should be split to `struct es10b_notification_metadata` and then that added as a member of `struct es10b_load_bound_profile_package_result` but it would be a bit of a backwards compatibility for `seqNumber` to move, so settled on joining in with extracting the metadata out.


maaarghk:
Force pushed 62b89ec to fix the clang-format check

septs (approved):
LGTM

Issue #209, opened 2025-03-08
Lpac with fibocom fm350-gl


mglants:
```
root@OpenWrt:~# lpac chip
AT_DEBUG: AT+CCHO=?
AT_DEBUG:
AT_DEBUG: OK
AT_DEBUG: AT+CCHC=?
AT_DEBUG:
AT_DEBUG: OK
AT_DEBUG: AT+CGLA=?
AT_DEBUG:
AT_DEBUG: OK
AT_DEBUG: AT+CCHC=1
AT_DEBUG:
AT_DEBUG: OK
AT_DEBUG: AT+CCHC=2
AT_DEBUG:
AT_DEBUG: ERROR
AT_DEBUG: AT+CCHC=3
AT_DEBUG:
AT_DEBUG: ERROR
AT_DEBUG: AT+CCHC=4
AT_DEBUG:
AT_DEBUG: ERROR
AT_DEBUG: AT+CCHO="A0000005591010FFFFFFFF8900000100"
AT_DEBUG:
AT_DEBUG: 1
AT_DEBUG:
AT_DEBUG: OK
{"type":"lpa","payload":{"code":-1,"message":"euicc_init","data":""}}
```
Could you help please what to do next?

SuperKali:
@mglants Use https://repo.superkali.me/releases/24.10.0/packages/additional_pack/lpac_2.1.1-r4_all.ipk

Issue #179, opened 2024-12-15
AT APDU issue without prefix response


stich86:
Hi @estkme,

me and my friends are playing with a Fibocom FM-350GL module that support eSIM provisioing using AT (it should works also over MBIM but not using USB mode). Now the big problem is that AT commands responds without put back the prefix of the called command, here is an example:

```
~# lpac chip info
AT_DEBUG: AT+CCHO=?
AT_DEBUG:
AT_DEBUG: OK
AT_DEBUG: AT+CCHC=?
AT_DEBUG:
AT_DEBUG: OK
AT_DEBUG: AT+CGLA=?
AT_DEBUG:
AT_DEBUG: OK
AT_DEBUG: T+CCHC=1
AT_DEBUG:
AT_DEBUG: OK
AT_DEBUG: AT+CCHO="A0000005591010FFFFFFFF8900000100"
AT_DEBUG: AT+CGLA=1,22,"81E2910006BF3E035C015A"
AT_DEBUG:
AT_DEBUG: 1
AT_DEBUG:
AT_DEBUG: OK

```

I was testing a modification to the loop in the close channel function, which currently counts up to 4. This module throws an error if the `AT+CCHC` command is sent without including 1.

In the test output, for the CGLA command, we see no reply as expected `(+CGLA: 1)`, but only `1`. A friend of mine reworked part of the AT PDU handling logic, but I think it can be improved further.

Here’s the repository for reference: https://github.com/SuperKali/lpac.

Do you have any suggestions? @SuperKali and I are happy to test any modifications you propose. Could you also help refine or correct the code?

Thanks in advance!

stich86:
@CoelacanthusHex are you able to help us? Thanks!

septs:
this depends on the at modem manufacturer's implementation

1alessandro1:
@septs did you even spent one minute to read what the issue is about?

SuperKali:
@1alessandro1 is right, it’s been two months since we opened this issue and it’s not acceptable to receive such an answer

stich86:
Hi @septs,

There is a fork of LPAC that uses the same commands and works.
The problem seems more related to how the module outputs the command response, but the various AT commands sent by LPAC to read/modify the eSIM are standard ones.

What we need to determine is whether it’s possible to add some checks on the output response (which is what the fork does). :)

septs:
@stich86 create a pr, plz

PR #427, opened 2026-03-27
fix(at_helpers): handle AT responses without command prefix (fix FM350-GL)

tomkosm:
## Summary                                                                                                                                                                                   
                                                                                                                                                                                               
  Fix `at_expect()` to handle AT responses without the standard `+CMD:` prefix.
                                         
  ## Problem                                                                                                                                                                                   
                                         
  The FM350-GL returns bare values instead of prefixed responses on AT+CCHO:

```
AT+CCHO="A0000005591010FFFFFFFF8900000100"
2
OK
```

  The parser expects `+CCHO: 2` and returns NULL, breaking channel open and APDU transmit.

  ## Fix

  Save the last non-echo content line during parsing. When OK is received but no prefix match was found, fall back to the saved line. Transparent to modems that return standard prefixed responses.


  Related: #179, #209

### DIFF
--- a/driver/apdu/at_cmd_unix.c
+++ b/driver/apdu/at_cmd_unix.c
@@ -57,6 +57,7 @@ int at_write_command(struct at_userdata *userdata, const char *command) {
 int at_expect(struct at_userdata *userdata, char **response, const char *expected) {
     char line[AT_BUFFER_SIZE];
     _cleanup_free_ char *found_response_data = NULL;
+    _cleanup_free_ char *last_content_line = NULL;
     int result = -1;
 
     if (response)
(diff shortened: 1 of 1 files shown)


septs (approved):
Wait for @CoelacanthusHex

driver/apdu/at_cmd_unix.c
```diff
@@ -126,12 +127,22 @@ int at_expect(struct at_userdata *userdata, char **response, const char *expecte
             found_response_data = strdup(line + strlen(expected));
             while (*found_response_data == ' ')
                 memmove(found_response_data, found_response_data + 1, strlen(found_response_data));
+        } else if (response && !found_response_data && strncasecmp(line, "AT", 2) != 0) {
+            // Fallback: save last non-echo content line for modems that return
+	    //	bare values without prefix (e.g. FM350-GL: "2\r\nOK" not "+CCHO: 2\r\nOK")
+            free(last_content_line);
+            last_content_line = strdup(line);
```
CoelacanthusHex: I think may be better to reuse variable found_response_data instead of introducing a new variable.

tomkosm: you are right fixed

CoelacanthusHex:
Please format the code and rebase on main branch to make CI happy.

PR #429, opened 2026-03-31
fix: handling of notification with sequence number 0 correctly

Eversberg:
Correctly check, if strtoul() converted the sequence number from command line to integer. A return value of 0 is not an error. Instead it successfully converted the command line argument to 0, unless the end pointer (str_end) equals the nptr.

This patch has been tested by processing and removing a notification with sequence number 0.


CoelacanthusHex (approved):
LGTM. The old code has a negative condition.

PR #435, opened 2026-05-10
Add support for Argos Mini II

fstanis:
This is a card reader that's very commonly used in some countries and I have one myself, so I analyzed the existing windows drivers and wrote this protocol based on that.


CoelacanthusHex:
Actually PCSCLite has support for this card reader[^1], you can use pcsc backend directly.

[^1]: https://github.com/LudovicRousseau/CCID/blob/fa26dbd42cdb7aa227ac9f1a08bcc5c5e0f0a32a/src/ccid_usb.c#L171

septs:
I don't think it's necessary to add this driver to implement it.

CoelacanthusHex:
Actually whatever USB smartcard reader (even other type reader like serial port card reader) should be implemented as a PCSC driver, and then use lpac PCSC backend, no need to implement a standalone driver in lpac. It's totally just re-invent a wheel. e.g. for your Argos Mini II (0b0c:0009), there is libagmiiusb project[^libagmiiusb]; for serial port NFC card reader like PN532, there is ifdnfc project[^ifdnfc].

[^libagmiiusb]: https://retis.santannapisa.it/~tommaso/libagmiiusb/
[^ifdnfc]: https://github.com/nfc-tools/ifdnfc

PR #439, opened 2026-06-14
cmake: add LPAC_CURL_GNUTLS for libcurl-gnutls HTTP backend

CAMOBAP:
## Summary

Adds an optional CMake switch to link the HTTP curl driver against **libcurl-gnutls** instead of the default **libcurl**.

`lpac profile download` talks HTTPS to SM-DP+ servers. On OpenWrt the default `libcurl` build uses mbedTLS and often fails certificate verification / TLS handshakes against LPA endpoints (**curl error 60**). OpenWrt ships a parallel `libcurl-gnutls` package (soname `libcurl-gnutls.so.4`) that works for SM-DP+ in practice.

This change lets packagers enable the alternate TLS backend at configure time without replacing `/usr/lib/libcurl.so.4` symlinks.

## Changes

- `cmake_dependent_option(LPAC_CURL_GNUTLS ...)` — default **OFF**, only relevant when `LPAC_WITH_HTTP_CURL=ON`
- `driver_http_curl` links `curl-gnutls` when enabled, otherwise unchanged `curl`

## Usage

```sh
cmake -DLPAC_CURL_GNUTLS=ON ...
```

OpenWrt feed packaging can expose this via Kconfig/Makefile (`-DLPAC_CURL_GNUTLS=ON`, `DEPENDS:=+libcurl-gnutls`).

## Related

- openwrt/packages#24963 — lpac + mbedTLS libcurl vs SM-DP+ HTTPS
- Verified on Fibocom FM350-GL (BPI-R4 Pro 8X, OpenWrt) with Truphone test profile download when linked against `libcurl-gnutls.so.4`

## Test plan

- [x] Default build (`LPAC_CURL_GNUTLS=OFF`) still links `curl`
- [x] `-DLPAC_CURL_GNUTLS=ON` links `curl-gnutls` and `lpac profile download` succeeds against `rsp.truphone.com` on OpenWrt with `libcurl-gnutls` installed

Made with [Cursor](https://cursor.com)


CoelacanthusHex:
I don't think it's the right way to fix this issue. This issue is because MbedTLS prior-3.6.0 doesn't support TLS 1.3 ciphersuites. I think it's better to add a check for TLS 1.3 support.

CAMOBAP:
@CoelacanthusHex thanks for the valuable insight!

Agree with your proposal, let me close this one

PR #444, opened 2026-07-13
euicc/es9p.c: Use snprintf to process untrusted data from the server

myrkr:
We process data here received in the JSON response from the server. This data is susceptible to MITM attacks as no TLS verification is performed and therefore we should not fully trust it.

By replacing the strncpy with snprintf we ensure that the output buffer is NUL-terminated even when an overflow has been prevented.


myrkr:
I applied the clang-format as suggested by one of the CI runs.

septs (approved):

CoelacanthusHex (approved):
LGTM, need review from @estkme

estkme:
My intuition tells me that cJSON automatically appends a null terminator when handling strings, but it never hurts to be cautious.

myrkr:
The problem is not missing terminating NUL in the input but the behaviour of strncpy which don't write a terminating NUL when truncation occurs on writing. strncpy should only be used if we can guarantee that the input is shorter than the destination buffer.

PR #442, opened 2026-06-14
driver/at: Fibocom FM350-GL compatibility

CAMOBAP:
## Summary

Three focused commits for the `main` branch AT driver (`at_cmd_unix.c`, `at_etsi.c`, `at_helpers.c`) so Fibocom FM350-GL eUICC provisioning works over the native `at` backend.

| Commit | Change |
|--------|--------|
| 1 | Bounded `poll()` deadlines in `at_expect_with_deadline()` (10s default, 240s CGLA for BPP) |
| 2 | Optional `AT+CCHO/CCHC/CGLA=?` capability probes (FM350 is silent) |
| 3 | Fibocom `AT+CCHO` response parsing (bare decimal channel id) + bounded CCHC recovery |

## Problem

Stock `main` AT driver fails on FM350-GL:

1. **Capability probes** (`AT+CCHO=?` etc.) — FM350 returns nothing; stock code treats this as fatal during `connect()`.
2. **`AT+CCHO`** — FM350 returns a bare decimal channel line, not `+CCHO: <id>` (related to #138-style session id handling).
3. **`AT+CGLA` during BPP** — profile download needs a wall-clock deadline beyond practical single-byte `VTIME` limits; infinite `poll()` can hang forever.

## Approach

Changes are **generic AT-driver robustness** improvements, not FM350-only `#ifdef`s:

- Probes use a short deadline and continue on silence (modems that answer `=?` with `OK` are unchanged).
- `at_expect_ccho_channel()` still accepts standard `+CCHO: <id>`; bare decimal lines are an additional accepted format.
- `at_expect_with_deadline()` replaces unbounded waits for all AT lines.

## Test evidence

**Platform:** OpenWrt on Banana Pi BPI-R4 Pro 8X  
**Modem:** Fibocom FM350-GL, FW `81600.0000.00.29.22.06`  
**AT port:** `/dev/ttyUSB3`, `AT+GTDUALSIM=1`, `AT+CMEE=0` before lpac

Built from `main` + this branch, default platform `libcurl` (mbedTLS on OpenWrt):

- [x] `lpac chip info` — EID read OK
- [x] `lpac profile list` — OK
- [x] Truphone BetterRoaming test profile BPP download — success (~26s)
- [x] Orange Flex production profile BPP download — success (~89s, separate session)

## Test plan for reviewers

- [ ] FM350-GL: `chip info`, `profile list`, profile download BPP
[...]

PR #446, opened 2026-07-19
driver/at: bounded poll deadlines in at_expect

CAMOBAP:
## Summary

First of 4 focused PRs replacing #442, split apart per review feedback there (thanks @CoelacanthusHex).

Replaces the unbounded `poll()` wait in `at_expect()` with a deadline-bounded `at_expect_with_deadline()`, so a silent/hung modem cannot block lpac forever on a single AT line.

- **Linux**: uses `timerfd_create()`/`timerfd_settime()` so the timeout is a real fd in the `poll()` set, instead of recomputing remaining time every wakeup (per [review feedback](https://github.com/estkme-group/lpac/pull/442#discussion_r3409801591)).
- **Other POSIX targets**: falls back to a poll+elapsed-time loop, since `timerfd` is Linux-specific and `at_cmd_unix.c` builds for any `UNIX` platform (e.g. macOS).
- `at_expect()` becomes a thin wrapper using `AT_DEFAULT_DEADLINE_MS`.
- **Win32**: delegates to the existing `at_expect()` until `COMMTIMEOUTS` are wired up (unchanged from before).

This is pure infrastructure — nothing calls `at_expect_with_deadline()` with a non-default deadline yet; that lands in the follow-up PRs.

## Test plan

- [x] Builds clean on macOS (fallback poll+elapsed path) and in a Linux container (timerfd path)
- [x] Fibocom FM350-GL: confirm no regression vs #442 test evidence (chip info / profile list / BPP download)
- [ ] Other AT modems (Sierra/Quectel): confirm unbounded-wait behavior change is transparent (10s default is generous for routine AT lines)

### DIFF
--- a/driver/apdu/at_cmd_unix.c
+++ b/driver/apdu/at_cmd_unix.c
@@ -16,6 +16,15 @@
 #include <unistd.h>
 
 #if defined(__linux__)
+#    include <sys/timerfd.h>
+#elif defined(__APPLE__) || defined(__FreeBSD__) || defined(__NetBSD__) || defined(__OpenBSD__) \
+    || defined(__DragonFly__)
+#    include <sys/event.h>
+#    include <sys/time.h>
+#    define AT_HAVE_KQUEUE 1
+#endif
+
+#if defined(__linux__)
 int enumerate_serial_device(cJSON *devices) {
     const char *dir_path = "/dev/serial/by-id";
     DIR *dir = opendir(dir_path);
@@ -54,7 +63,52 @@ int at_write_command(struct at_userdata *userdata, const char *command) {
     return 0;
 }
 
-int at_expect(struct at_userdata *userdata, char **response, const char *expected) {
+/* Extracts the next complete line from `buf` into `line`, decoupled from at_userdata.
+ * Returns 0 on success, -1 if no complete line is buffered yet. */
+static int at_pop_line(char *buf, size_t *buf_len, char *line, size_t line_size) {
+    char *newline = memchr(buf, '\n', *buf_len);
+    if (!newline)
+        return -1;
+
+    size_t line_len = newline - buf;
+    if (line_len >= line_size)
+        line_len = line_size - 1;
+    memcpy(line, buf, line_len);
+    line[line_len] = '\0';
+
+    memmove(buf, newline + 1, *buf_len - line_len - 1);
+    *buf_len -= line_len + 1;
+
+    line[strcspn(line, "\r")] = 0;
+    return 0;
+}
+
+/* Applies AT response bookkeeping to an already-extracted `line`.
+ * Returns 1 to keep reading, 0 once `result` is final. */
+static int at_match_line(const char *line, const char *expected, char **found_response_data, int *result) {
+    if (strlen(line) == 0)
+        return 1;
+
+    if (strcmp(line, "ERROR") == 0) {
+        *result = -1;
+        return 0;
+    }
+    if (strcmp(line, "OK") == 0) {
+        *result = 0;
+        return 0;
+    }
+
+    if (expected && strncmp(line, expected, strlen(expected)) == 0) {
+        free(*found_response_data);
+        *found_response_data = strdup(line + strlen(expected));
+        while (*found_response_data && (*found_response_data)[0] == ' ')
+            memmove(*found_response_data, *found_response_data + 1, strlen(*found_response_data));
+    }
+    return 1;
+}
+
+#if defined(__linux__)
+int at_expect_with_deadline(struct at_userdata *userdata, char **response, const char *expected, int deadline_ms) {
     char line[AT_BUFFER_SIZE];
     _cleanup_free_ char *found_response_data = NULL;
     int result = -1;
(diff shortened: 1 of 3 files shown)


CoelacanthusHex:
> * **Other POSIX targets**: falls back to a poll+elapsed-time loop, since `timerfd` is Linux-specific and `at_cmd_unix.c` builds for any `UNIX` platform (e.g. macOS).

FYI,
1. FreeBSD has native support for timerfd since FreeBSD 14, before that, they support timerfd via linux compatibility layer. NetBSD and Illumos (The successor of OpenSolaris) also support timerfd.
2. All BSD (include macOS) has support for `kqueue` (which is a general event system and can be used as better replacement of `poll`), which has an event named EVFILT_TIMER.

itay-sho:
Thanks again for the contribution, @CAMOBAP!
As I mentioned on the original combined PR, I was able to test this successfully on a Fibocom FM350-GL running in USB mode on a Banana Pi R4.

Please let me know if any further checks or testing are needed on my end to help move this PR forward. Since the changes are now split across separate PRs, testing full workflows directly might be a bit more involved, but I'm happy to give it a shot if you have a recommended approach.

One thing I noticed during my testing and wanted to ask about:

I managed to download profiles from two different operators using the original merged PR. During that process, I noticed the notifications queue (for reporting events like profile installation success or removal back to the SM-DP+ server) gets populated, but it doesn't seem to automatically flush/push those notifications.

Unless triggered manually (e.g., via lpac), the notifications remain queued indefinitely. I stumbled upon this because one operator was waiting for the SM-DP+ confirmation before activating the line, which didn't happen until I manually triggered the push.

Is this manual push the intended behavior, or should notification processing happen automatically as part of the profile installation/deletion flow?

CAMOBAP:
> Is this manual push the intended behavior, or should notification processing happen automatically as part of the profile installation/deletion flow?

To be honest IDK. From what I understand this behavior wasn't touched by my changes 

Maybe @CoelacanthusHex can bring some broader context

CoelacanthusHex:
> Unless triggered manually (e.g., via lpac), the notifications remain queued indefinitely. I stumbled upon this because one operator was waiting for the SM-DP+ confirmation before activating the line, which didn't happen until I manually triggered the push.
> 
> Is this manual push the intended behavior, or should notification processing happen automatically as part of the profile installation/deletion flow?

It's intended behavior. lpac is a low-level tool mainly for developers so it won't do something which may be convenient but shouldn't be done in some situations.

CoelacanthusHex (commented):
I suggest adding a kqueue implementation to cover most BSDs.

driver/apdu/at_cmd_unix.c
```diff
@@ -54,78 +63,250 @@ int at_write_command(struct at_userdata *userdata, const char *command) {
     return 0;
 }
 
-int at_expect(struct at_userdata *userdata, char **response, const char *expected) {
+/* Consumes one line from userdata->at_read_buffer into `line` and applies the common
+ * AT response bookkeeping (OK/ERROR detection, `expected` prefix capture).
+ * Returns 2 if no complete line is buffered yet (caller should wait for more data),
+ * 1 if a line was processed and the caller should check for another one, or 0 once
+ * `result`/`found_response_data` are final. */
```
CoelacanthusHex, line 70:
```suggestion
/*
 * Consumes one line from userdata->at_read_buffer into `line` and applies the common
 * AT response bookkeeping (OK/ERROR detection, `expected` prefix capture).
 * Returns 2 if no complete line is buffered yet (caller should wait for more data),
 * 1 if a line was processed and the caller should check for another one, or 0 once
 * `result`/`found_response_data` are final.
 */
```
Align comment format with other code.

CoelacanthusHex, line 70: And this function has design issue: it mixed logic part and IO part. Uncompleted line should be handled by caller instead of a logic functon. You should only passed a buffer with at least one complete line into at_process_line().

CAMOBAP: Addressed both, thanks for the review:

- Comment format aligned with the suggested `/*\n * ...\n */` block style.
- Split `at_process_line()` into `at_pop_line()` (pure I/O buffering — finds/extracts/strips a line, knows nothing about AT semantics) and `at_match_line()` (pure logic — OK/ERROR detection, `expected` prefix capture, no I/O or buffer access). The caller now only calls `at_match_line()` once it already has a complete line from `at_pop_line()`, so the two concerns are no longer mixed.

Verified with a standalone functional harness (socketpair-based, exercising the timerfd/kqueue/poll+elapsed paths directly) covering data-ready, timeout, delayed-arrival, and multi-line-burst-in-one-read cases — all pass on both the Linux and macOS code paths.

CAMOBAP:
Added the `kqueue()`/`EVFILT_TIMER` implementation for macOS/FreeBSD/NetBSD/OpenBSD/DragonFly (previously they fell through to the generic poll+elapsed-time fallback). Structurally mirrors the Linux timerfd branch: registers `EVFILT_READ` on the fd plus a one-shot `EVFILT_TIMER` for the deadline, blocks on a single `kevent()` call, and distinguishes timeout vs. data-ready by which filter fired. No `NOTE_MSECONDS` flag needed since the timer's data field defaults to milliseconds without a `NOTE_*SECONDS` flag on every kqueue implementation, including older BSDs.

Verified functionally (not just compiled) via a socketpair-based harness driving `at_expect_with_deadline()` directly — data-ready, timeout, delayed-arrival, and multi-line-burst cases all pass on the kqueue path.

driver/apdu/at_cmd_unix.c
```diff
@@ -54,78 +63,262 @@ int at_write_command(struct at_userdata *userdata, const char *command) {
     return 0;
 }
 
-int at_expect(struct at_userdata *userdata, char **response, const char *expected) {
+/*
+ * Extracts the next complete line from userdata->at_read_buffer into `line`
+ * (CR-stripped, NUL-terminated) and removes it from the buffer. Pure I/O
+ * buffering: it knows nothing about AT response semantics.
+ * Returns 1 if a line was extracted, 0 if the buffer doesn't contain a
```
CoelacanthusHex: I prefer to use 0 for successful extract and -1 for a failed extract to follow C language tradition.

driver/apdu/at_cmd_unix.c
```diff
@@ -54,78 +63,262 @@ int at_write_command(struct at_userdata *userdata, const char *command) {
...
+ * complete line yet (caller should read more data before calling again).
+ */
+static int at_pop_line(struct at_userdata *userdata, char *line, size_t line_size) {
```
CoelacanthusHex: I prefer to make this function accept a buffer and buffer size to make it not coupled with at_userdata.

driver/apdu/at_cmd_unix.c
```diff
@@ -54,78 +63,262 @@ int at_write_command(struct at_userdata *userdata, const char *command) {
...
+
+    line[strcspn(line, "\r")] = 0;
+    return 1;
+}
+
+/*
+ * Applies the AT response bookkeeping (OK/ERROR detection, `expected` prefix
+ * capture) to an already-extracted `line`. Pure logic: no I/O, no buffer
+ * access.
+ * Returns 1 if the caller should keep reading (blank or unmatched line), or 0
```
CoelacanthusHex: Use 0 and -1 as well.

CoelacanthusHex (commented):
Some comments are unnecessary; please write comments that are concise and to the point.

driver/apdu/at_cmd_unix.c
CAMOBAP: Fixed — `at_pop_line()` now returns 0 on success / -1 when no complete line is buffered yet.

driver/apdu/at_cmd_unix.c
CAMOBAP: Fixed — `at_pop_line()` now takes `(char *buf, size_t *buf_len, char *line, size_t line_size)` instead of `struct at_userdata *`, so it's no longer coupled to it.

driver/apdu/at_cmd_unix.c
CAMOBAP: Fixed, same change as above.

CAMOBAP:
Addressed all of the latest round:

- \`at_pop_line()\` now returns 0 on success / -1 when no line is buffered yet (was 1/0), and takes `(char *buf, size_t *buf_len, char *line, size_t line_size)` instead of `struct at_userdata *`, so it's decoupled from it.
- Trimmed the doc comments on `at_pop_line()`/`at_match_line()` and the kqueue `NOTE_MSECONDS` note down to one line each.

Re-verified with the same functional harness (data-ready/timeout/delayed-arrival/multi-line-burst) on the Linux, kqueue, and generic poll+elapsed paths — all still pass with the new convention.

septs:
I oppose any further improvements to the AT driver implementation.

https://github.com/estkme-group/lpac/blob/3ff35594ec15062a3ed10c3da1c26eb0a13390b8/docs/backends/at.md?plain=1#L3-L11

CAMOBAP:
@septs ynderstood on the "demo purposes only" framing, but I'd push back on calling this scope creep specifically for the timeout part (kqueue/FM350 parsing are separate discussions).

Right now `at_expect()` calls `poll(&pfd, 1, -1)` — an infinite wait. If a modem goes silent mid-command, `lpac` hangs forever with zero recourse — no error, no way out short of `kill -9`. That's not "insufficient response time causing a rejected request" (which the doc already warns about), it's a hang with no exit. I'd call that a real defect in the existing implementation, not a nice-to-have improvement on top of it.

Given the concern though — happy to make the default behavior unchanged: keep infinite wait as the default, and make the deadline an opt-in env var (same pattern as `AT_DEBUG`/`AT_DEVICE` already in this file) for anyone who wants the bounded behavior. That way nobody's existing setup changes unless they explicitly ask for it, and the hang is still fixable for people who hit it. Would that be an acceptable middle ground?

PR #452, opened 2026-07-22
driver/http: make TLS certificate/hostname verification configurable

sch-m:
By default lpac's HTTP backends do not verify the SM-DP+/SM-DS
server's TLS certificate or hostname, since some servers use
certificates issued by CAs that are not (yet) part of the GSMA CI
trust store shipped with common OS trust stores. This makes it
impossible to opt back into full verification even when the target
server's certificate is trusted.

Add an environment variable to opt back into standard TLS
verification:

  LPAC_HTTP_SSL_VERIFY=1

When unset or falsy, behavior is unchanged for both backends
(verification disabled), preserving backward compatibility.

curl backend (driver/http/curl.c):
- When enabled, both CURLOPT_SSL_VERIFYPEER and CURLOPT_SSL_VERIFYHOST
  are set to enable verification (CURLOPT_SSL_VERIFYHOST is set to 2,
  matching curl's requirement that this option only accepts 0 or 2).
- The setting is read once in libhttpinterface_init() and applied to
  every request made via http_interface_transmit().

WinHTTP backend (driver/http/winhttp.c):
- When enabled, WINHTTP_OPTION_SECURITY_FLAGS is set to 0 instead of
  SECURITY_FLAG_IGNORE_UNKNOWN_CA, so the request goes through
  WinHTTP's standard certificate verification.
- Same as the curl backend, the setting is read once in
  libhttpinterface_init() and applied via a static variable in
  http_interface_transmit().


CoelacanthusHex:
IMO, I will want to expand it to WinHTTP backend as well and named it as LPAC_HTTP_SSL_VERIFY.

sch-m:
> IMO, I will want to expand it to WinHTTP backend as well and named it as LPAC_HTTP_SSL_VERIFY.

OK, I've renamed the option and added WinHTTP backend changes (untested) as well.

driver/http/curl.c
```diff
@@ -6,6 +6,16 @@
...
 
+/*
+ * By default lpac does not verify the SM-DP+/SM-DS server's TLS certificate
+ * or hostname, since some servers use certificates issued by CAs that are
+ * not (yet) part of the GSMA CI trust store shipped with common OS trust
+ * stores. Setting this environment variable to a truthy value re-enables
+ * both checks. This is intentionally not backend-specific (see
+ * HTTP_ENV_NAME), so all HTTP backends that support it use the same name.
+ */
+#define ENV_SSL_VERIFY "LPAC_HTTP_SSL_VERIFY"
```
septs, line 17:
```suggestion
#define ENV_SSL_VERIFY ENV_HTTP_DRIVER "_SSL_VERIFY"
```

https://github.com/estkme-group/lpac/blob/26597de5797be011424fc14cea96ec4e00855a71/utils/lpac/utils.h#L15

sch-m: Thanks for the hint. I've changed the define for both curl and winhttp backend.

PR #453, opened 2026-07-23
fix(profile/discovery): parse every command-line option

tsaQB:
The profile discovery option loop calls `getopt()` in the `while` condition and again at the end of each iteration. The second call consumes every other option without passing it through the `switch`.

For example:

```text
lpac profile discovery -s smds.example.test -i 490154203237518
```

The `-s` option is processed, while `-i` is consumed by the redundant call and ignored.


septs (approved):

CoelacanthusHex (approved):
LGTM

PR #457, opened 2026-09-12
fix(euicc): return the padded length from hexutil_gsmbcd2bin so even-length (18-digit) ICCIDs keep their 0xFF padding

adaasch:
**Title:**
`fix(euicc): return the padded length from hexutil_gsmbcd2bin so even-length (18-digit) ICCIDs keep their 0xFF padding`

---

> **Disclaimer:** This fix was produced with the help of agentic AI.
> However, every step was supervised by me, and the code and tests were
> verified manually.

**Summary**

`es10c_enable/disable/delete_profile()` currently emit a 9-byte `5A` (Iccid)
TLV for even-length (e.g. 18-digit) ICCIDs. SGP.22 requires
`Iccid ::= [APPLICATION 26] OCTET STRING (SIZE(10))` (rsp.asn line 23), and
the eUICC compares the request against the 10-byte ICCID it reported in
GetProfilesInfo — so all three operations fail with `iccidOrAidNotFound`
(result = 1) for exactly those profiles. 19-digit ICCIDs are unaffected: the
'F' pad nibble produced by the odd-nibble conversion is part of the converted
bytes, so the TLV is already 10 bytes.

**Motivation — who is affected**

At least German o2 (Telefónica Deutschland, including MVNOs on their
network) issues 18-digit ICCIDs — exactly the even-length class this bug
hits. A profile from o2 therefore cannot be enabled, disabled or deleted
by any lpac-master-based LPA (OpenEUICC, EasyEUICC, …), while profiles
with 19-digit ICCIDs on the same card keep working. Because the failure
depends only on the ICCID length of that one profile, it looks like a
profile-specific glitch: "only this profile fails, the others are fine"
— which is how it surfaced in the downstream report. Any other carrier
with even-length ICCIDs would be equally affected.

**Root cause** (one line)

`euicc_hexutil_gsmbcd2bin()` pads the output buffer with 0xFF up to
`padding_to` but returns the *pre-padding* byte count; the ICCID call site in
`es10c_enable_disable_delete_profile()` uses that return value as the TLV
length, so the padding is written and then dropped.

**Regression history**

- 57e5e31 (2024-05, "[Not-Tested] Fix for 18-digit iccid") returned the
  **padded** length — wire-correct. All tagged releases through v2.3.0
  (2025-08-15) are fine.
- 57c175d (#323, 2025-09-18) rewrote the helper and switched to
  `return bytes`, introducing both this regression and the then-unguarded
  memset overflow for the IMEI caller.
- 80c8268 (#415, 2026-01-31) correctly guarded the overflow but kept
  `return bytes`, so the ICCID regression survived. All master builds since
  2025-09-18 are affected; no tagged release is.
- Downstream, OpenEUICC picked this up with its lpac master bump on
  2026-03-08 (pin `d214738`); see the user-facing bug report at
  <https://gitea.angry.im/PeterCxy/OpenEUICC/issues/332>.

**The fix**

Return `max(bytes, padding_to)`. The IMEI caller in `es10b.c` passes
`padding_to = 0` and is therefore unchanged (and stays free of the overflow
fixed by #415). `es10c_set_nickname()` is unaffected (explicit
`sizeof(asn1iccid)` = 10).

**Evidence** (ICCID digits masked; byte lengths and FF padding preserved)

1. *Live repro, stock OpenEUICC ecosystem (downstream).* With the official,
   unmodified EasyEUICC v1.7.2 release (Verbose Logging), disabling an
   18-digit o2 profile on a Sm@rtSIM eUICC sends:

       DisableProfile:  BF3210 A00B 5A09 98xxxxxxxxxxxxxx42 8101 FF
       response:        BF3203 8001 01   (iccidOrAidNotFound), retry 8101 00 → same

   while GetProfilesInfo in the same session returns

       5A0A 98xxxxxxxxxxxxxx42 FF      (10 bytes)

   (Samsung Galaxy J7, removable Sm@rtSIM, ARA-M-authorized cert; full log
   available in the linked OpenEUICC issue discussion.)

2. *A/B with this patch, independent third client (lpac CLI over PC/SC,
   same chip, built from OpenEUICC's pin `d214738` vs. this branch):*

       stock:  BF3210 A00B 5A09 98xxxxxxxxxxxxxx42    810100  →  BF3203 8001 01  FAIL
       fixed:  BF3211 A00C 5A0A 98xxxxxxxxxxxxxx42FF  810100  →  BF3203 8001 00  OK
       fixed:  BF3111 A00C 5A0A 98xxxxxxxxxxxxxx42FF  810100  →  BF3103 8001 00  OK (re-enabled)

   The fixed build's ICCID is byte-identical to the card's GetProfilesInfo
   form, and the card accepts it for both directions.

3. *Host-level check* (`padding_to=10`): 18-digit input returns 10 after the
   patch (was 9), encodes `5A0A …FF`; 19-digit input unchanged (`5A0A …F5`);
   IMEI path (`padding_to=0`) unchanged, no overflow.

**Suggested regression test**

An 18-digit ICCID must encode to a 10-byte `5A` value equal to
(digits + `FF`). Happy to add to a hexutil test target if you point me at
the preferred place (no existing hexutil test target found in-tree).

---


septs (approved):

septs:
LGTM, @CoelacanthusHex please review

euicc/hexutil.c
```diff
@@ -95,7 +95,7 @@ int euicc_hexutil_gsmbcd2bin(uint8_t *restrict output, const uint32_t output_len
     if (padding_to > (uint32_t)bytes) {
```
CoelacanthusHex, line 95: Why not merge these two condition branch? e.g. set `bytes = padding_to` in if branch.

septs: `return padding_to` in the branch is best?

adaasch: It would be identical. If you like the early return more, I am happy to change that.

adaasch: Changed to early return.

septs (approved):

CoelacanthusHex:
Now we have unit tests for lpac. Please rebase on main and add regression tests for this issue.

adaasch:
Anything else to do here? Or just wait?

PR #461, opened 2026-09-19
fix(pkgconfig): use the configured libeuicc libdir

samcday:
Fedora uses ${prefix}/lib64 rather than ${prefix}/lib.

Use the CMAKE_INSTALL_FULL_LIBDIR variable from GNUInstallDirs module instead for maximum distro compatibility.


CoelacanthusHex (approved):
LGTM

